PR #4, opened 2019-12-10
Add support for the VF interface type

steven-webster:
StarlingX now supports the concept of a VF interface, which is a
sub-interface of a regular ethernet SR-IOV interface.  The VF
interface typically uses a subset of the parent interface's
virtual functions which can be assigned to a  different VF driver
and data network.

Note: depends on corresponding gophercloud modification https://github.com/Wind-River/gophercloud/pull/2

Signed-off-by: Steven Webster <steven.webster@windriver.com>


steven-webster:
This pull request has been updated to grab the merged VF interface type gophercloud mods (update to GoPkg.lock), as well as updated charts and code generated by the tool versions specified in Developer.md

config/crds/starlingx_v1_host.yaml
```diff
@@ -436,6 +436,86 @@ spec:
...
+                            type: string
+                          mtu:
+                            description: MTU defines the maximum transmit unit for
+                              this interface.
+                            format: int64
+                            maximum: 9216
+                            minimum: 576
+                            type: integer
+                          name:
+                            description: SystemName defines the name of the interface
```
albailey: Not that it matters but SystemName is not the field here.

albailey:
Seems fine to me

config/crds/starlingx_v1_host.yaml
steven-webster: Thanks Al.  I've refreshed the pull request. Seems it was an existing typo in hostprofile_types.go ->CommonInterfaceInfo that affected the description of all interface types.  There's a similar issue for VolumeGroupInfo and EthernetPortInfo.  I can fix those up in a followup PR, as they are not directly associated with this one.

PR #5, opened 2019-12-12
Make bmType set to none by default

donpenney:
The bm_type field in starlingx sysinv database is set to none by
default, so do the same in deployment manager. If not specified, set
the value to none rather than nil.


config/crds/starlingx_v1_host.yaml
```diff
@@ -63,6 +63,15 @@ spec:
                     address:
                       description: Address defines the board management IP address.
                       type: string
+                    type:
```
albailey: I will assume that since this passed your testing,  the field name "type"  does not conflict with the "type" reserved field

pkg/controller/host/host_controller.go
```diff
@@ -1183,7 +1188,13 @@ func (r *ReconcileHost) ReconcileExistingHost(client *gophercloud.ServiceClient,
 	// build a true representation of the system state.  Trying to reconcile
 	// the system generated BMC info will always result in an error because
 	// it does not contain password info.
-	defaults.BoardManagement = nil
+	bmType := "none"
+	bmInfo := starlingxv1.BMInfo{
+		Type:        &bmType,
```
albailey: I assume this syntax is fine  (the & looks odd to me)

steven-webster:
LGTM

pkg/apis/starlingx/v1/zz_generated.deepequal.go
```diff
@@ -898,6 +898,14 @@ func (in *MatchBMInfo) DeepEqual(other *MatchBMInfo) bool {
 		}
 	}
 
+	if (in.Type == nil) != (other.Type == nil) {
+		return false
+	} else if in.Type != nil {
```
mpeters-wrs: Shouldn't this need to check that other.Type is not nil otherwise it could generate an error.  I know this follows the other fields, but I'm not sure why it doesn't also have the possibility of dereferencing a nil value.

donpenney: The very first check in the function is:
        if other == nil {
                return false
        }

mpeters-wrs: It is the dereference of *other.Type that I am concerned about.  I don't see that checked anywhere.

donpenney: Isn't that caught by the check on line 901? If one is nil and one is not, then it returns false. So if in.Type is non-nil at 903, then other.Type must also be non-nil

mpeters-wrs: Yes.  Sorry, you are correct.

PR #13, opened 2020-02-18
Add support for configuring storage backends

sdinescu-wrs:
Changes included in this commit:
- added a reconcile function for storage backends
- updated examples to include storage backend information

Notes:
- this only supports adding storage backends (not editing or
  deleting)
- It only supports adding the default Ceph storage backend;
  External ceph support is not included


albailey:
looks good to me.

PR #14, opened 2020-02-19
Fix ServiceParameter creation for optional resource and personality

albailey:
The resource and personality fields are not required when creating
a new service parameter. They are optional, and needed to be declared
omitempty in system_types, and trigger the CRD generated files to
have the correct validation.

Signed-off-by: albailey <Al.Bailey@windriver.com>


steven-webster:
+1 LGTM

PR #15, opened 2020-02-19
Fix node deletion issue when node is already deleted from system

donpenney:
When a node has been deleted from the system prior to deleting the
resource, the deployment-manager gets stuck due to a bug in the
reconciler. This update resolves the bug by clearing retrieved data in
order to follow the correct code path.


albailey:
lgtm

PR #16, opened 2020-02-21
Add version info to deploy tool

donpenney:
This update builds in version information to the deploy tool,
collecting data from git in the Makefile and passing into the go build
as ldflags.


steven-webster:
LGTM

PR #18, opened 2020-02-22
Include tool version info in deploy config output

donpenney:
This update adds tool version info, along with the generation
timestamp, to the config yaml file generated by the deploy tool in a
comment block.


donpenney:
Added a comment to version.go to indicate variables are set by the Makefile

PR #19, opened 2020-02-24
Fix service parameters blocking system resource from going insync

albailey:
The system service parameters were not merging with the spec, so the
insync was never completing.

Also, the personality and resource fields were omitted, so any new
service parameters that added a personality or a resource would
not go insync either.

Signed-off-by: albailey <Al.Bailey@windriver.com>


steven-webster:
LGTM:  Should probably bump the copyright date on these 3 files

PR #24, opened 2020-03-10
Update certificate reconciliation

donpenney:
This update adds an IsKeyEqual method for CertificateInfo and updates
the reconciler to allow for proper syncing when the deployment config
includes a platform certificate.

### DIFF
--- a/pkg/controller/system/system_controller.go
+++ b/pkg/controller/system/system_controller.go
@@ -874,7 +874,11 @@ func (r *ReconcileSystem) ReconcileCertificates(client *gophercloud.ServiceClien
 				return err
 			}
 
-			log.Info(fmt.Sprintf("skipping %q certificate %q from system", c.Type, c.Secret))
+			// If we don't find the corresponding secret, this is most likely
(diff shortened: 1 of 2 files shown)


pkg/controller/system/system_controller.go
```diff
@@ -874,9 +874,8 @@ func (r *ReconcileSystem) ReconcileCertificates(client *gophercloud.ServiceClien
 				return err
 			}
 
-			msg := fmt.Sprintf("waiting for %q certificate %q to be created", c.Type, c.Secret)
-			r.WarningEvent(instance, common.ResourceDependency, msg)
-			return common.NewMissingKubernetesResource(msg)
+			log.Info(fmt.Sprintf("skipping %q certificate %q from system", c.Type, c.Secret))
```
mpeters-wrs: I think we need to state why this dependency is no longer being enforced.  In general it means we can now have a misconfigured Certificate entry pointing to a non-existent Secret.  Likely this needs to be more than just an Info log since it could also be a misconfiguration.  So keep the event, and log a warning.

PR #25, opened 2020-03-19
Fix replicationFactor for Ceph backend

sdinescu-wrs:
The input for replicationFactor was int, but the backend
expected string. The simple fix is to just do the transformation
in out code.

Also changed the minimum acceptable value for replicationFactor
from 2 to 1, since AIO-SX has a default replication of 1.


steven-webster:
Looks good

PR #31, opened 2020-04-01
Add support for storage networks

sdinescu-wrs:
Changes included in this commit:
- add storage as a supported network type
- fix incorrect use of network ID instead of address pool ID
- fix incorrect comparison to determine  address allocation type


pkg/build/build.go
```diff
@@ -463,21 +464,35 @@ func (db *DeploymentBuilder) buildPlatformNetworks(d *Deployment) error {
...
 
 	nets := make([]*starlingxv1.PlatformNetwork, 0)
+	always_generate_networks := []string{"storage"}
+	sort.Strings(always_generate_networks)
 	for _, p := range pools {
 		skip := false
+		network_type := networks.NetworkTypeOther
 		for _, n := range results {
 			if n.PoolUUID == p.ID {
 				// TODO(alegacy): for now we only support networks used for data
```
albailey, line 474: Is this original comment still valid?

sdinescu-wrs: Yes, for all other platform networks. I just checked specifically for  storage type networks. That is why I added a new comment to describe this special case, while also leaving the original comment intact.

PR #37, opened 2020-06-11
Remove filesystem validation enum

mwang3-wrs:
Deployment Manager would fail if dc-vault filesystem was added to
the deployment configuration file even though it is supported. The
filesystem validation enum was removed to avoid similar future
incidences. Invalid filesystem configurations are handled by sysinv.


config/crds/starlingx_v1_system.yaml
```diff
@@ -297,15 +297,6 @@ spec:
...
                           resource.
-                        enum:
-                        - cgcs
-                        - platform
-                        - cinder
-                        - database
-                        - docker-distribution
-                        - etcd
-                        - extension
-                        - patch-vault
```
mpeters-wrs, line 308: These currently line up with the set defined in gophercloud / controllerFilesystems.  Should we consider adjusting that list to only include the "well-known" file systems that need special handling?

PR #43, opened 2020-07-13
Fix IP validation for IPv6 addresses

mwang3-wrs:
Deployment manager would fail when IPv6 addresses in the deployment
config contained leading zeros. To fix this issue, a transformation was
applied to all addresses when read in from the model to remove leading
zeros. Error checking was added for invalid IP addresses.


mwang3-wrs:
After further experimenting, I found that the changes are only effective for address fields under the Host section. I think I need to apply the same transformations to gateway and subnet fields

PR #44, opened 2020-08-04
Add support for multiple CA certificates

mwang3-wrs:
This change allows multiple CA certificates to be installed by
specifying them in the deployment config.

### DIFF
--- a/pkg/apis/starlingx/v1/system_types.go
+++ b/pkg/apis/starlingx/v1/system_types.go
@@ -57,7 +57,7 @@ type CertificateInfo struct {
 // point to a Secret named by the system.
 func (in *CertificateInfo) DeepEqual(other *CertificateInfo) bool {
 	if other != nil {
-		return in.Type == other.Type
+		return (in.Type == other.Type) && (in.Secret == other.Secret)
 	}
 
 	return false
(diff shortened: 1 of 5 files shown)


pkg/controller/system/system_controller.go
```diff
@@ -848,8 +848,9 @@ func (r *ReconcileSystem) PrivateKeyTranmissionAllowed(client *gophercloud.Servi
 // ReconcileCertificates configures the system certificates to align with the
 // desired list of certificates.
 func (r *ReconcileSystem) ReconcileCertificates(client *gophercloud.ServiceClient, instance *starlingxv1.System, spec *starlingxv1.SystemSpec, info *v1info.SystemInfo) error {
+
```
donpenney, line 851: The blank line seems out of place. Was this added unintentionally?

mwang3-wrs: Yes, I'll remove that.

pkg/apis/starlingx/v1/system_types.go
```diff
@@ -57,7 +57,7 @@ type CertificateInfo struct {
 // point to a Secret named by the system.
 func (in *CertificateInfo) DeepEqual(other *CertificateInfo) bool {
 	if other != nil {
-		return in.Type == other.Type
+		return (in.Type == other.Type) && (in.Secret == other.Secret)
```
donpenney, line 60: Per discussion, it appears the Secret value from the model does not align with the Secret value from sysinv, so this field would not be able to be used for comparing items

mwang3-wrs:
Gophercloud changes: https://github.com/Wind-River/gophercloud/pull/7

PR #46, opened 2020-08-28
Fix for provisioning TPM certs

mwang3-wrs:
This update ensures that certificates configured for TPM are installed
after the host is unlocked and available. Common code for processing
certificate contents is refactored.


mwang3-wrs:
For now, it's best to live with the limitation of only installing TPM certificates via CLI/API

PR #47, opened 2020-09-11
Add retry loop in Makefile

mwang3-wrs:
Retry "go test" up to 3 times in case of encountering an intermittent issue
where the control plane fails to start


donpenney (approved):
lgtm

albailey (approved):
looks good to me

PR #48, opened 2020-09-17
Remove controller-0 restrictions to support reinstall

donpenney:
This update includes the following:
1. Update nodeSelector value to allow DM pod to run on either
controller
2. Remove restriction that controller-0 must use dynamic provisioning,
to provide more flexibility when reinstalling controller-0
3. Updating provisioningIsAllowed check to support adding nodes when
either controller is unlocked-enabled
4. Remove tolerances from values.yaml allowing DM pod to run on
locked controller (post initial config. VIM is not running to set the
taint until after initial unlock)


pkg/controller/host/host_controller.go
```diff
@@ -208,7 +208,7 @@ func (r *ReconcileHost) buildInitialHostOpts(instance *starlingxv1.Host, profile
 
 func provisioningAllowed(objects []hosts.Host) bool {
 	for _, host := range objects {
-		if host.Hostname == hosts.Controller0 {
+		if host.Hostname == hosts.Controller0 || host.Hostname == "controller-1" {
```
mpeters-wrs, line 211: Is the plan to create a PR for gophercloud longer term?

donpenney: If anything, I'd rather drop the Controller0 constant from gophercloud and create constants locally within the DM repo, since it doesn't appear to be used anywhere in gophercloud

mpeters-wrs: I agree.  This is not an API constant but rather a well know resource of StarlingX and therefore doesn't belong in the gophercloud API.

mwang3-wrs (approved):
lgtm

mpeters-wrs (approved):

PR #51, opened 2021-01-28
Fix helm init issue for builder

donpenney:
The "helm init" command in the Dockerfile.builder is failing due to
the relocation of the stable repo. This updates the command to
specify the new location with the --stable-repo-url option to use
https://charts.helm.sh/stable


albailey (approved):
approved

mpeters-wrs (approved):

PR #58, opened 2021-02-09
Add retries to ssl certificate check

5abeelwr:
When DM tries to install ssl certificate, instead of
reporting error immediately if secret is not found, we
retry and log messages before reporting errors.

Signed-off-by: 5abeelwr <sabeel.ansari@windriver.com>

### DIFF
--- a/pkg/controller/system/system_controller.go
+++ b/pkg/controller/system/system_controller.go
@@ -48,6 +48,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 	"strconv"
 	"strings"
+	"time"
 )
 
 var log = logf.Log.WithName("controller").WithName("system")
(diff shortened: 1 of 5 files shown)


donpenney (approved):
LGTM

pkg/controller/system/system_controller.go
```diff
@@ -150,13 +151,18 @@ func (r *ReconcileSystem) installRootCertificates(instance *starlingxv1.System)
...
+		for iter := 0; iter < numRetries; iter++ {
+			caBytes, ok = secret.Data[starlingxv1.SecretCaCertKey]
+			if !ok {
+				log.Info("Platform certificate CA not ready/available", "name", c.Secret)
+				time.Sleep(5 * time.Second)
+			} else {
+				log.Info("Platform certificate CA ready!")
+				break
+			}
 		}
```
mwang3-wrs, line 166: Should there be a log to inform the user that DM will continue without a CA certificate if it times out?

mwang3-wrs (approved):

PR #59, opened 2021-02-11
Remove SNMP host based logic from DM

jinfanzon:
Since SNMP is containerized and no longer runs host-based,
then all SNMP related logic from DM has to be removed in
order to avoid errors.

NOTE: SNMP will be present in WRCP 20.06 unless patchback to remove SNMP is applied.


jinfanzon:
This pull request on branch 2.0.5 will be closed since if no patchback(a patch that removes SNMP host-based functionality) is applied, it should be working as it is, else if applied community and trap destination APIs will return empty values so no errors will be raised.

PR #60, opened 2021-02-11
Remove SNMP host based logic from DM

jinfanzon:
Since SNMP is containerized and no longer runs host-based,
then all SNMP related logic from DM has to be removed in
order to avoid errors.


jinfanzon:
This PR is closed since the commit message is erroneous

PR #63, opened 2021-02-19
Remove SNMP temporary fix

jinfanzon:
The temporary fix for avoiding SNMP error that occured when SNMP
was not host-based was removed, since SNMP removal patchback
will keep SNMP host-based APIs, but with empty functionality.
Then if patchback is not applied, then behavior remains.


jinfanzon:
Closed per discussion with @donpenney

PR #66, opened 2021-03-01
Add new example yaml files

gaolitao:
Add new example yaml files for single nic feature and rate limiting configuraiton.

    - aio-sx-single-nic.yaml
    - aio-sx-vf-rate-limiting.yaml


donpenney (changes requested):
Please look at the infrastructure in place for building example models. These are not just copies of a system's config model. There is a README file that explains the build.

The Makefile builds the example files, which can then be added to a commit:
https://github.com/Wind-River/cloud-platform-deployment-manager/blob/master/Makefile#L140-L151

donpenney (approved):
LGTM. Thanks!

mwang3-wrs (approved):

steven-webster (commented):
LGTM, thanks Litao!

PR #67, opened 2021-03-18
Introducing GEO location coordinates to system

dpintobarros:
Two new fields was added to System Spec.
The fields latitude and longitude was added to store GEO
location, so the deployment manager can set the system's
GEO location coordinates following the
starlingx_v1_system.yaml config.

Signed-off-by: Daniel Pinto Barros <DanielPinto.Barros@windriver.com>


donpenney:
The gophercloud changes must be made separately first:
https://github.com/Wind-River/cloud-platform-deployment-manager/blob/master/DEVELOPER.md#working-with-a-private-fork-of-a-vendor-package

You would need to fork the wind-river/gophercloud repo and make these changes in the "starlingx" branch, to be pulled into your DM code via the "dep ensure -update" command described in the DEVELOPER.md page. Once that change is posted in a pull request, and approved and merged, you'd then need to update the DM PR again with an updated reference to the GC SHA.

https://github.com/Wind-River/gophercloud/tree/starlingx

PR #68, opened 2021-03-22
Introducing GEO location coordinates to Deployment Manager

dpintobarros:
New fields was added to System Spec.
The fields latitude and longitude was added to system spec,
so the deployment manager can set the GEO location coordinates
following the starlingx_v1_system.yaml config.
Gophercloud was updated.

Signed-off-by: Daniel Pinto Barros <DanielPinto.Barros@windriver.com>


pkg/apis/starlingx/v1/system_types_test.go
```diff
@@ -20,6 +20,8 @@ func TestStorageSystem(t *testing.T) {
 	description := string("A sample description")
 	location := string("A sample location")
 	contact := string("A sample contact")
+	latitude := string("A sample latitude")
+	longitude := string("A sample longitude")
```
donpenney, line 24: Shouldn't these be somewhat valid latitude/longitude values?

pkg/apis/starlingx/v1/system_types.go
```diff
@@ -265,6 +265,20 @@ type SystemSpec struct {
 	// +optional
 	Location *string `json:"location,omitempty"`
 
+	// Latitude is the latitude geolocation coordinate of the system's physical
+	// location.
+	// +kubebuilder:validation:Pattern=^[a-zA-Z0-9\-_\. ]+$
```
donpenney, line 270: There's no restrictions on what this string should be?

donpenney: Ok, there's a comment on the StarlingX review from Matt Peters that the format has been specifically left generic:
https://review.opendev.org/c/starlingx/config/+/776670/14#message-d0bb4b7fb9c9fa04ba09fc9b11364b6ff7c49666

pkg/apis/starlingx/v1/system_types_test.go
dpintobarros: I think that there is no difference for testing, but I can set these sample to valid values

donpenney: Thanks. That would serve as a bit of an example, at least, short of adding something under the examples dir

examples/aio-sx.yaml
```diff
@@ -164,6 +164,8 @@ spec:
   contact: info@windriver.com
   description: Virtual Box Standard System
   location: vbox
+  latitude: 45.35189954974955
+  longitude: -75.91866628453701
```
donpenney, line 168: If you're going to add values to an example model, please read the README file in the examples dir. These files are not modified directly. Rather, there are "kustomize" subdirectories under examples with segments of files, and the example models get generated by the build.

donpenney (approved):
LGTM

albailey (approved):
seems fine to me

examples/aio-sx.yaml
mwang3-wrs: Given Don's comment, should segment be moved to its own yaml file in another directory dedicated for this GEO location feature?

donpenney: That's certainly an option, creating a new aio-sx-geo-location.yaml file from the Makefile, which would clearly show this as an optional setting.

donpenney: Please update per Melissa's suggestion, creating a new geo-location subdir under the examples/aio-sx dir and a new aio-sx-geo-location.yaml generated by the Makefile, rather than adding it to the default. I'm surprised, actually, that this isn't added to other aio-sx example files, since you'd added it to the default config.

For example of adding new subsets, see this recent update that added two:
https://github.com/Wind-River/cloud-platform-deployment-manager/commit/f07a9bcf88d5a3f496d7992310a5b77685319eb9

donpenney (approved):

examples/aio-sx/geo-location/geo-location.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: starlingx.windriver.com/v1
+kind: System
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: vbox
+  namespace: geo-location
```
donpenney, line 7: Why is this a separate namespace?

examples/aio-sx/geo-location/geo-location.yaml
```diff
@@ -0,0 +1,10 @@
...
+spec:
+  latitude: 45.35189954974955
+  longitude: -75.91866628453701
```
mwang3-wrs, line 10: I'm not sure the lack of a newline at the end of a file will impact the generation of the yaml file but it might be safer to add it in.

examples/aio-sx/geo-location/geo-location.yaml
dpintobarros: It was the way I tested it here, but I can reset namespace to deployment and the name to geo-location

examples/aio-sx-geo-location.yaml
```diff
@@ -0,0 +1,185 @@
...
+  subfunctions:
+  - controller
+  - worker
+---
+apiVersion: starlingx.windriver.com/v1
+kind: System
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: geo-location
```
donpenney, line 161: The file you're generating is putting your latitude/longitude tags in a separate section from the other System because your "name" doesn't match that of the default Section, similar to how you previously had a unique namespace. Please update the name to align with the default Section "vbox", regenerate the file, and ensure your latitude and longitude settings are contained in the same System section as other settings.

donpenney (changes requested):

examples/aio-sx/geo-location/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+bases:
+  - ../default
+
+patches:
+  - geo-location.yaml
```
donpenney, line 5: You fixed the missing EOL on geo-location.yaml, but not this one.

donpenney (approved):

mwang3-wrs (approved):

PR #69, opened 2021-03-24
Add bond option primary_reselect

teresa-ho:
This update is to allow the option primary_reselect configurable for
aggregated ethernet interface. The option is to prevent reverting
between the primary slave and other slaves.

Signed-off-by: Teresa Ho <teresa.ho@windriver.com>


donpenney (commented):
LGTM. There will need to be an update once the corresponding gophercloud update is merged, to update the SHA reference

donpenney (approved):
LGTM

PR #73, opened 2021-05-13
Add retries for applying the deployment-config

mwang3-wrs:
In environments where it may take longer for DM to restart, the 10
second timeout is not enough when waiting for DM to be ready. This
change updates the playbook to retry with delays when applying the
deployment-config.

### DIFF
--- a/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
+++ b/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
@@ -107,5 +107,9 @@
 
         - name: Apply Deployment Configuration File
           shell: KUBECONFIG=/etc/kubernetes/admin.conf /bin/kubectl apply -f deployment-config.yaml
+          register: apply_deployment_config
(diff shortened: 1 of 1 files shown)


docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -107,5 +107,9 @@
 
         - name: Apply Deployment Configuration File
           shell: KUBECONFIG=/etc/kubernetes/admin.conf /bin/kubectl apply -f deployment-config.yaml
+          register: apply_deployment_config
+          retries: 5
+          delay: 10
+          until: apply_deployment_config.rc == 0
```
donpenney, line 113: Presumably, you could get rid of the 10 second delay that occurs just ahead of this now, letting the first apply fail and allowing the retries to handle it, but I think this is fine and is consistent with current behaviour

donpenney (approved):

PR #74, opened 2021-05-18
Copy default-registry-key into DM namespace

donpenney:
In order to support pulling images from the private registry.local, if
the images are not already in the containerd cache, the secrets must
be populated in the platform-deployment-manager namespace by copying
the default-registry-key after DM has been installed.


mwang3-wrs (approved):

albailey (commented):
You have this command  kubectl --kubeconfig=/etc/kubernetes/admin.conf   but also the KUBECONFIG variable

albailey (approved):

PR #79, opened 2021-06-10
Fix DM reinstall error in helmv2

mwang3-wrs:
When DM exists in the helmv2 namespace but helmv3 is used, the
command "helm upgrade --install" will fail. The playbook would try
to install DM in the helmv3 namespace which conflicts with the
existing release in the helmv2 namespace. This change checks which
version of helm the DM release is installed in and uses a separate
reinstallation procedure for helmv2.


donpenney (approved):
lgtm

albailey (approved):

PR #81, opened 2021-09-03
Update DM playbook to use a fixed path

mwang3-wrs:
Previously the deployment charts and overrides files had to be
available in the user's home directory. This change allows any
fixed path to be specified in the ansible overrides when running
this playbook.


albailey:
this shows as work in progress and unable to merge ?

albailey (approved):
Seems gine to me

PR #91, opened 2022-02-23
Add validation webhooks for PTP configuration

mwang3-wrs:
This change ensures that the format for the parameters in
ptpinterfaces and ptpinstances is valid when applying the
deployment configuration file.


yjian118 (commented):
nit: all the licenses should be started from 2022.

Can you mention the test process in the commit message?

yjian118 (approved):

albailey (approved):

PR #92, opened 2022-02-24
Accept multiple PTPInterface association

yjian118:
Previously, we only accept one PTP interface associated with an
interface. This commit enables the interface can accept multiple PTP
interface association.

Test:
1 Modified the deployment configuration, assigned two PTP interfaces
to an interface.
2 Bootstrap a AIOSX, deploy this node with the DM and modified
configuration
3 After the deployment, the AIOSX can be correctly deployed and
unlocked.
4 Check the PTP interfaces, both of them are assigned to a single
interface.


config/crds/starlingx_v1_host.yaml
```diff
@@ -343,12 +343,16 @@ spec:
...
-                            description: PtpInterface defines the ptp interface to
-                              be configured against this interface.
+                          ptpInterfaces:
+                            description: PtpInterfaces defines the ptp interfaces
+                              to be configured against this interface.
+                            items:
+                              maxLength: 255
+                              pattern: ^[a-zA-Z0-9\-_]+$
+                              type: string
                             maxLength: 255
```
albailey, line 353: I assume we are fine with 255 being the max size for the array.

albailey (approved):

mwang3-wrs (approved):

PR #101, opened 2022-04-26
Bug fix: Support more than 2 storage nodes

ttakenak:
The issue was the second pair of storage nodes are
not configured if more than 2 storage nodes are
specified (Unlock fails)

The fix is adding a new state "RetryCephPrimaryGroupNotReady"
to wait to add second group of storage nodes until the first group
of storage nodes are configured as unlocked and available.

Tested by configure deployment nodes with designer DM container:
- SX (Not to have regression issues)
- DX (Not to have regression issues)
- Storage (4 Storage nodes, confirm storage-2 and storage-3 waited
           until storage-0 and storage-1 are available then both
           nodes are configured as expected)

### DIFF
--- a/pkg/controller/host/host_controller.go
+++ b/pkg/controller/host/host_controller.go
@@ -10,6 +10,7 @@ import (
 	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/hosts"
 	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/labels"
 	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/ptpinstances"
+	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/storagebackends"
 	perrors "github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


pkg/controller/host/host_controller.go
```diff
@@ -1519,6 +1520,94 @@ func (r *ReconcileHost) ReconcileResource(client *gophercloud.ServiceClient, ins
...
+	} else {
+		return n, err
+	}
+
+	log.Info("Ceph replication factor", "num", n)
+	return n, nil
+}
+
+// Function to judge the specified host is in the ceph
+// primary group or not. If host name is start with
```
yjian118, line 1542: We don't have the naming convention to assign the storage nodes as "storage-0", "storage-1" etc. So this check cannot use the host name. I'll recommend to use the host id and its personality to do this check

ttakenak: I removed host name to identify the primary storage node group and update the patch.
Thank you for your review.

yjian118 (approved):
Looks good to me

albailey (approved):

PR #103, opened 2022-05-06
Migrate helm CRDs from apiversion v1beta1 to v1

KaustubhDhokte:
This change migrates all the helm chart CRDs from
apiextensions.k8s.io/v1beta1 to apiextensions.k8s.io/v1.

The apiextensions.k8s.io/v1beta1 API version of CRD is not
supported from k8s version 1.22.5.
https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/

The manifests need to be updated to use apiextensions.k8s.io/v1

Manifests are updated according to following deprecation guide
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#migrate-to-non-deprecated-apis

Note:
According to https://www.pulumi.com/registry/packages/kubernetes/api-docs/apiextensions/v1/customresourcedefinition/
'served' and 'storage' are mandatory fields of
CustomResourceDefinitionVersion.
With 'served=true', this version can be served via REST
APIs. 'storage=true' signifies that this version should be
used to persist custom resources to the storage.

Test Plan:
PASS: helm chart deployment is successful.
FAIL: platform-deployment-manager pod is Running
      Reason: 'manager' container fails to start
      This change only includes CRD apiversion change.
      We will need to migrate Webhook resources similar
      to this.

Signed-off-by: Kaustubh Dhokte <kaustubh.dhokte@windriver.com>

### DIFF
--- a/config/crds/starlingx_v1_datanetwork.yaml
+++ b/config/crds/starlingx_v1_datanetwork.yaml
@@ -1,129 +1,135 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   creationTimestamp: null
   labels:
-    controller-tools.k8s.io: "1.0"
+    controller-tools.k8s.io: '1.0'
   name: datanetworks.starlingx.windriver.com
 spec:
-  additionalPrinterColumns:
-  - JSONPath: .spec.type
-    description: The data network encapsulation type.
-    name: type
-    type: string
-  - JSONPath: .status.inSync
-    description: The current synchronization state.
-    name: insync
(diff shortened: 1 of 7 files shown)


yjian118:
I must admit the scope of this commit is beyond my estimation, but all the CRDs in DM should be auto-generated by the kubebuilder during the make, modifying the YAMLs is not acceptable. And we have many dependencies relying on this API, see: https://github.com/Wind-River/cloud-platform-deployment-manager/blob/master/vendor/sigs.k8s.io/controller-tools/pkg/internal/codegen/parse/crd.go#L57.  I'm going to scan the repo to see how can we update it

yjian118:
Close this PR as it is no longer applied

PR #108, opened 2022-06-01
Retry retrieve secret in webhook

ttakenak:
-Secret is created but when the moment webhook is retrieve by
 name, it is possible that secret is invisible in context.
 This fix is retry to retrieve secret.
-Later system is installing certifaction in /etc but it
 failed since the manager runs in nonroot user
 (This is default by kubebuilderv3).
 This fix is running in root as the current DM.

TEST PLAN:
PASS: Install DM and apply configuration, which had error to obtain secret in webhook, in bare metal environmrnt.
           Finished successfully to apply configuration and host status becomes "unlocked-enabled-available".


yjian118:
The user permission is identical with what we have in kubebuilder v1: https://github.com/Wind-River/cloud-platform-deployment-manager/blob/stable/final_ver_kubebuilder_v1/Dockerfile#L22 and accessing that directory requires root permission, looks good to me

yjian118 (approved):

albailey (approved):

PR #109, opened 2022-06-13
Correct copyright

yjian118:
This commit corrects the copyrights in the source code with the
SPDX format which Wind River is using.


ttakenak:
It looks good to me.

ttakenak (approved):

PR #112, opened 2022-07-14
Update gophercloud with changed attribute datatypes

hmatsui-encora:
This commit updates gophercloud module reference to an updated commit
that aligns some attributes datatypes between the API and database.

Reference: https://github.com/Wind-River/gophercloud/commit/aac06588b172f270f9bd62fc38ebd627c6e4a895

This commit should merge along with this one, otherwise DM will fail:
Depends-on: https://review.opendev.org/c/starlingx/config/+/848026

Test Plan:
PASS: Build an image and use it to configure and unlock an AIO-SX

Signed-off-by: Heitor Matsui <HeitorVieira.Matsui@windriver.com>


go.sum
```diff
@@ -470,6 +470,8 @@ github.com/tmc/grpc-websocket-proxy v0.0.0-20190109142713-0ad062ec5ee5/go.mod h1
 github.com/tmc/grpc-websocket-proxy v0.0.0-20201229170055-e5319fda7802/go.mod h1:ncp9v5uamzpCO7NfCPTXjqaC+bZgJeR0sMTm6dMHP7U=
 github.com/wind-river/gophercloud v0.0.0-20220503140232-192cf245ebfd h1:SqpOmGVa4q5yHaILQKtLJ1W2/uZuL6Rt3moDK4gXodo=
```
yjian118, line 471: I think we may need to remove the reference of the previous gophercloud commit here: v0.0.0-20220503140232-192cf245ebfd, can you remove that and try to make the image?

yjian118 (changes requested):
I checked the whole project including the CRDs and controllers, but cannot find the code relating to the "VM1GHugepagesEnabled" and "VM2MHugepagesEnabled". So despite the previous commit of gophercloud, I'm okay with this pull request.

go.sum
hmatsui-encora: It works, I updated the commit, thanks for the review

albailey (approved):

PR #114, opened 2022-07-26
Migrate golangci-lint

yjian118:
https://install.goreleaser.com/github.com/golangci/golangci-lint.sh is not valid any more.
Need to update the DEVELOPER.md and Dockerfile.builder which refer it.
This commit opts https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh v1.13
for the stable/v2.0.8 branch. This commit also contains some change to fit for this golangci-lint.

Regarding the master branch, as we are using go v1.17, we'd better to choose a later version
of golangci-lint, and make changes to adapt the requirement of the new go version.

Test: Created new image with "make" and "make docker-build"

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -98,7 +98,10 @@ project now uses the GoLangCI-lint tool which is also in use by newer
 Kubebuilder versions.
 
 ```bash
-curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.17.1
+# We used this link to download the golangci-lint, but it is not valid any more
+#curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.17.1
+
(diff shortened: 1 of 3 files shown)


hugonicodemos (approved):

ttakenak:
Can we also modify Makefile to run target "golangci"?
Originally it was;
----------
# Run go vet against code
vet: golangci
	go vet ./pkg/... ./cmd/...
----------
(I removed this dependency "golangci" from target "vet" during migration to kubebuilder V3 base...)

yjian118:
> Can we also modify Makefile to run target "golangci"? Originally it was;
> 
> ```
> # Run go vet against code
> vet: golangci
> 	go vet ./pkg/... ./cmd/...
> ```
> 
> (I removed this dependency "golangci" from target "vet" during migration to kubebuilder V3 base...)

@ttakenak I think you are referring to the make file in the master branch. This commit is against the v2.0.8 branch which we will expect a bug fix for this stable version. As I mentioned in the commit message, I expect to choose a later version of the golangci-lint which will be fit for go v1.17. And I believe there may be more work to do to adapt the new golangci-lint. May need to plan more effort on the master branch

ttakenak:
@yjian118 , Sorry, my mistake, I thought your change would be in master.

ttakenak (approved):

PR #117, opened 2022-07-29
Test code migration

ttakenak:
When we migrated DM from v1 to v3, we left test code migration
because of the limited time.
This is test code migration from the kubebuilder v1 base to v3 base.

kubebuilder v3 created the test template with ginkgo and gomega.
So that the existing test code is modified to adapt to this environment
but no new testing scenario is added in this patch

Test Plan:
PASS: Run "make test" and finished successfully
PASS: Run "make && DEBUG=yes make docker-build" and finished successfully
PASS: Confirm test coverage is similar as before migration

### DIFF
--- a/api/v1/constructors_test.go
+++ b/api/v1/constructors_test.go
@@ -1,0 +1,97 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+/* Copyright(c) 2019-2022 Wind River Systems, Inc. */
+
+package v1
+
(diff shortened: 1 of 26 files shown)


yjian118 (approved):
This change looks good to me. Would expect a following commit regarding the makefile and the installation of the ginkgo if it is required for other contributors

enzocandotti (approved):

PR #122, opened 2022-08-23
Workaround for StringList and pointer in deepcopy

ttakenak:
Currently controller-tools, which generate deepcopy code
from type, has the following problems:
1. No support validating internal list items on list types
https://github.com/kubernetes-sigs/controller-tools/issues/342
2. Invalid deepcopy generation with nested pointer to type-aliased slices
https://github.com/kubernetes-sigs/controller-tools/issues/586
Both issue reports have stayed open more than a year.

This fix is doing workaround to generate deepcopy properly.
1. Set a alias of the item type to introduce the validation
   (This is for StringList)
2. Replace array to list type after generated deepcopy code

Test Plan:
PASS: Run "make && DEBUG=yes make docker-build" and finished successfully
PASS: Run "make generate" and generated deepcopy and deepequal
      and no compile error in both generated code
PASS: Run "make manifests" and confirm validation is reflected to CRDs
PASS: Install SX successfully with designer DM and helm charts


yjian118 (approved):

hugonicodemos (approved):
LGTM, thanks takamasa.

PR #123, opened 2022-08-24
Fix make golangci error

ttakenak:
It had error because golangci-lint version was old.
Now using the latest version of golangci-lint(v1.49.0)

Modify Makefile to download golangci-lint automatically,
so that user does not need to install golangci-lint
manually.
golangci-lint will be placed [local copy of repo]/bin
directory.

Fix some lint erros.

Test Plan:
PASS: "make golangci" finish successfully and no lint error
PASS: "make golangci-lint" downloaded the specified version of
      golangci-lint and placed into local bin directory
PASS: "make golangci-lint" does not download golangci-lint
      again, if golangci-lint exists in local bin directory
PASS: "make && DEBUG=yes make docker-build" finish successfully
PASS: Install SX successfully with designer DM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ helm/wind-river-cloud-platform-deployment-manager/templates/rbac
 *.swp
 *.swo
 *~
+.vscode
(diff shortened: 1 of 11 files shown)


yjian118 (changes requested):
I would expect to correct the reference of the golang ci in: https://github.com/Wind-River/cloud-platform-deployment-manager/blob/master/Dockerfile.builder#L16 and the contributor.md in this PR.

yjian118 (approved):

hugonicodemos (approved):

PR #140, opened 2022-11-29
Upversion to 2.0.10

enzocandotti:
Signed-off-by: Enzo Candotti <enzocandotti93@gmail.com>


ttakenak (commented):
Can we leave DEVELOPER.md as it is because it is a description that we use v3 since v2.0.9.

ttakenak (approved):

yjian118 (approved):

PR #143, opened 2022-12-05
Deploy tool inserts invalid MaxCPUMhzConfigured entry 

MahalakshmiManthiram:
Deploy tool inserts invalid MaxCPUMhzConfigured entry in deployment-config file.  After adding null check for MaxCPUMhzConfigured, deploy tool doesn't made invalid entry. When MaxCPUMhzConfigured have some value, deploy tool persists that value.

Test Plan:

PASS: "make && DEBUG=yes" and "make docker-build" finished successfully 
PASS: Tested the code changes with newly built chart and added the value for MaxCPUMhzConfigured field in deployment-config.yaml file and ran deploy tool and value was perisisted in config file 
PASS: Tested the code changes with newly built chart and with null value in MaxCPUMhzConfigured, Deploy tool doesnt generate invalid entry in deployment-config.file

Signed-off-by: mmanthir <mahalakshmi.manthiram@windriver.com>


api/v1/constructors.go
```diff
@@ -801,7 +801,11 @@ func NewHostProfileSpec(host v1info.HostInfo) (*HostProfileSpec, error) {
 	}
 	spec.Console = &host.Console
 	spec.InstallOutput = &host.InstallOutput
-	spec.MaxCPUMhzConfigured = &host.MaxCPUMhzConfigured
+
+	if host.MaxCPUMhzConfigured != "" {
```
hugonicodemos, line 805: Should we need to verify if not nil here?

MahalakshmiManthiram: Yes, to avoid invalid entry in config file

hugonicodemos (approved):

yjian118 (approved):

PR #170, opened 2023-01-18
Fix device path if it is offered as device node

yjian118:
This commit fix the issue if the device path is offered as device node in the deployment configurations.

Test plan
1 . Deploy a storage system with the root_device, boot_device are configured as /dev/sda, the cgts-lv is configured as /dev/sda5, the osd is configured as /dev/sdb in the controller-0
2. Varify the system is deployed with the device path associated with the device node. All resource are in-sync and reconciled after the deployment.

### DIFF
--- a/controllers/host/host_controller.go
+++ b/controllers/host/host_controller.go
@@ -1,5 +1,5 @@
 /* SPDX-License-Identifier: Apache-2.0 */
-/* Copyright(c) 2019-2022 Wind River Systems, Inc. */
+/* Copyright(c) 2019-2023 Wind River Systems, Inc. */
 
 package host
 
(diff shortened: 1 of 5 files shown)


controllers/host/host_controller.go
```diff
@@ -1328,6 +1326,11 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
 		return err
 	}
 
+	// Fix attributes in profiles to a uniformed format
+	// As the Merge Profiles will overwrite some formate in the default profile
```
albailey, line 1330: typo formate

hugonicodemos (approved):

albailey (approved):

PR #180, opened 2023-02-17
DM upgrade support

sselvara1:
We need to consider the removal of webhook configuration, webhook service and webhook secret before DM upgrade. For this, check webhook configuration, webhook service and webhook secret and remove them if they exist before apply DM in ansible.

Test Plan:
PASS: "make && DEBUG=yes make docker-build" finish successfully PASS: Upgrade DM from 2.0.6 to 2.0.8 after system upgrade
      - Ansible finish successfully
      - No unexpected error message found in DM log
PASS: Run DM ansible multiple times successfully
PASS: Apply configuration multiple times without
      unexpected errors
      (DM does not support day-2 configuration yet but
       it should not show unexpected errors)
PASS: Install SX successfully with designer DM

### DIFF
--- a/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
+++ b/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
@@ -164,16 +164,92 @@
 
       when: helmv2_dm_exists.rc == 0
 
+    # Remove v1 webhook, webhook service & webhook secret
+    - block:
+      - name: Search webhook configurations for v1
+        shell: |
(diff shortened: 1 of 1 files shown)


yjian118:
Regarding the "without unexpected errors", have you checked if there's any error shown in the dm manager container's log? Have you checked if the secret is newly created after the playbook?

sselvara1:
There is no error shown in dm manager container's log.
The secret is newly created after the playbook.

Below is the snapshots - 

image

image

yjian118 (approved):

albailey (approved):

PR #182, opened 2023-02-24
Upversion to 2.0.10-1

sselvara1:
This commit upversions to 2.0.10-1


sselvara1:
> I would prefer to close this commit as we normally don't do up version in the master branch, but I'm open to discussion as the content in this pull request is fine

Agree, here is the PR to up version on the stable/v2.0.10 branch,
https://github.com/Wind-River/cloud-platform-deployment-manager/pull/183

PR #192, opened 2023-03-14
Upversion to 2.0.10-2

yjian118:
This commit upversions to 2.0.10-2


api/v1/constructors.go
sselvara1, line 1: how did it work previously with ParseVolumeGroupInfo ? and now its parseVolumeGroupInfo

DEBUGGING.md
sselvara1, line 1: up version to be done for every commit we are going add here onwards?

yjian118: Yes... That's the conclusion we get with the registry admin

api/v1/constructors.go
yjian118: Would prefer to start with uppercase if it is a public method. But this change is not included in this pull request

DEBUGGING.md
yjian118: To be more precise, the registry admin won't do the image tag update and re-push work, so if an image is pushed to the staging area, we can only update the image tag if we want more commits to be included

ttakenak (approved):

PR #200, opened 2023-03-24
HwSettle: Fixed default value and validation annotation.

sselvara1:
This commit updates format of the default value in examples and the regular expression for validation annotation.


yjian118:
Need to rebase the commit and resolve merge conflicts

examples/aio-sx-hw-settle.yaml
```diff
@@ -96,7 +96,7 @@ spec:
   administrativeState: unlocked
   bootDevice: /dev/disk/by-path/pci-0000:00:0d.0-ata-1.0
   console: tty0
-  hwSettle: 0
+  hwSettle: "0"
```
yjian118, line 99: Would prefer to use a non-zero value in the example

yjian118:
Based on: https://github.com/Wind-River/cloud-platform-deployment-manager/blob/stable/master/api/v1/constructors.go#L808-L810, and  https://github.com/Wind-River/cloud-platform-deployment-manager/blob/stable/v2.0.10/api/v1/hostprofile_types.go#L860. I need to know if the hw_settle will be generated as "0", would prefer it to be omitted if it is zero value and keep the CRD as-is.

config/crd/bases/starlingx.windriver.com_hostprofiles.yaml
```diff
@@ -202,7 +202,7 @@ spec:
               hwSettle:
                 description: HwSettle defines the wait time for SCSI devices to show
                   up.
-                pattern: ^[1-9][0-9]*$
```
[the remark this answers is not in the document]
sselvara1: we need to set hwSettle: "0" - this is the default value.
deploy tool generates hwSettle: "0". Not hwSettle: ""

examples/aio-sx-hw-settle.yaml
sselvara1: As discussed, default value is "0"

hugonicodemos (approved):
LGTM

yjian118 (approved):
Just a reminder to patch it back to the release stable branch

PR #206, opened 2023-05-03
Add DeploymentScope state to resources

enzocandotti:
This commit adds a new state to the host, platformnetwork, ptpinstance, ptpinterface and system CRD's. The state can have a value of either 'bootstrap' or 'principal', depending on whether the resource was configured on initial deployment or during a Day 2 update. The default value is 'bootstrap', and in future commits the state update will be added in order to support Day-2 operations.

The purpose of this new state is to provide clearer resource status information and to detect out-of-sync resources during an update, which will trigger an alarm.

Test Plan

    PASS: make and build the new image. Install Deployment Manager on
    a system using this image and describe the affected CRDs. Verify
    that the new state is added to the 'Status' section as:

    Deployment Scope:
                    Default:      bootstrap
                    Description:  DeploymentScope defines whether the
                                  resource has been deployed on the
                                  initial setup or during an update.
                    Enum:
                      bootstrap
                      principal
                    Type:  string
    PASS: use get_namespaced_custom_object() method from the K8s API to
    monitor the resources. Verify that the new state is present on the
    response, and its default value is 'bootstrap'.


enzocandotti:
Hi all,
The "DeploymentType" name can be changed, it's just a suggestion. Feel free to suggest a new name.

This state is currently not visible in the resource information when running 'kubectl get'. If you think it would be useful to display it, please let me know and I can add it.
Thanks!

hugonicodemos (approved):
LGTM. Please add the Test Plan section in the commit message.

ttakenak:
Can we add "printcolumn" annotation so that we can see this new status with the command "kubectl get".

enzocandotti:
> Can we add "printcolumn" annotation so that we can see this new status with the command "kubectl get".
Done.

> Based on Matt's request, he was using the "scope" argument with "bootstrap" and "principle" options, is there any concern to align with those terms?
Done. The naming now is DeploymentScope: [bootstrap/principal]

Thanks @ttakenak @yjian118

enzocandotti:
Example of kubectl get... :

`[sysadmin@controller-0 ~(keystone_admin)]$ kubectl -n deployment get datanetworks,hosts,platformnetworks,systems,ptpinstances,ptpinterface
NAME                                               TYPE   INSYNC   RECONCILED   SCOPE
datanetwork.starlingx.windriver.com/group0-data0   vlan   true     true         bootstrap
datanetwork.starlingx.windriver.com/group0-data1   vlan   true     true         bootstrap

NAME                                        ADMINISTRATIVE   OPERATIONAL   AVAILABILITY   PROFILE                INSYNC   RECONCILED   SCOPE
host.starlingx.windriver.com/controller-0   unlocked         enabled       available      controller-0-profile   true     true         bootstrap

NAME                                  MODE      TYPE         VERSION   INSYNC   RECONCILED   SCOPE
system.starlingx.windriver.com/vbox   simplex   all-in-one   22.12     false    true         bootstrap
`

ttakenak (approved):

PR #214, opened 2023-05-17
Manage reconciler configuration trigger by deployment scope

ttakenak:
Annotation "last-applied-configuration" is updated depend on
deployment scope, so that
  - Reconcile can be triggered if the scope is principle after reconciled
  - Reconcile can be blocked if the scope is changed to bootstrap after reconciled

Here is how to set instance status and annotation by deployment scope:
If deploymentScope: "principal" in conf
  - Set SCOPE="principal" in instance status
  - Add annotation deployment-manager/reconcile-after-insync: true

If deploymentScope: "bootstrap" or no deploymentScope or empty value in conf
  - Set SCOPE="bootstrap" in instance status
  - Delete annotation deployment-manager/reconcile-after-insync

If deploymetnScope with any other values than above, it returns error.

Test Plan:
PASS: Fresh install with designer DM
PASS: "make && DEBUG=yes make docker-build" finish successfully
For following each reconciler, Test:
  - Reconcile can be triggered if the scope is principle after reconciled
  - Reconcile can be blocked if the scope is changed to bootstrap after reconciled

PASS system reconciler
PASS host reconciler
PASS platform network reconciler
PASS data network reconciler
PASS ptp instance reconciler
PASS ptp interface reconciler


yjian118:
The format of the 1st section commit message is confusing to me. Can you update it with proper indentation?

yjian118 (approved):

ttakenak:
I think PP message is formatted automatically. I formatted the message. But message in commit is fine.

PR #215, opened 2023-05-19
Ignore OpenLDAP certificate in system resource

enzocandotti:
The OpenLDAP certificate is installed during the initial unlock, so it cannot be collected after the bootstrap. This is causeing the system resource to be insync=false after unlock.

This commit intends to ignore this certificate in the DM, so DM will not take action on it.

Test Plan:
PASS: Build an image with these changes. Install it on a system. Verify that after the initial unlock the type: openldap certificate is not present on the current spec.


ttakenak (approved):

yjian118:
Expect a test: using the deploy tool to generate the deployment-config.yaml to see whether the cert will be there

PR #218, opened 2023-06-01
Addition of deployctl phoney in Makefile

sharbhar:


sselvara1 (commented):
description need to be added.

hugonicodemos:
> description need to be added.

+1

Update the commit message with a description and the test plan section.

yjian118 (changes requested):
What's the problem with go vet?

sharbhar:
> What's the problem with go vet?

During go vet run
golangci-lint run ./controllers/... is executed .

We are getting type-check error
image


Else we excluded.

yjian118:
I see, it doesn't make sense to run "golangci-lint run ./controllers/.. " when building the deployctl. But creating two phonies to build the deployctl is definitely wrong. 

Can you make the change like this:
1 move the golangci-lint out of go vet, create a new phony golangci 
2 append the golangci phony to the existing phonies with vet except the tools

yjian118 (changes requested):
I can not agree to have two phonies to reach a same goal

sharbhar:
> I see, it doesn't make sense to run "golangci-lint run ./controllers/.. " when building the deployctl. But creating two phonies to build the deployctl is definitely wrong.
> 
> Can you make the change like this: 1 move the golangci-lint out of go vet, create a new phony golangci 2 append the golangci phony to the existing phonies with vet except the tools

The changes are done.
Please review.

PR #223, opened 2023-06-08
Fix count ceph primary group

ttakenak:
This fix provides the proper way to judge ceph primary group is ready or not.

Test Plan:
PASS: Fresh install storage
PASS: Fresh install storage with delay unlock one node in
      the secondary group
PASS: Fresh install storage with delay configuration one
      node in the secondary group


enzocandotti (approved):
LGTM

yjian118 (approved):

PR #226, opened 2023-06-16
Add Delta status between spec vs current config

enzocandotti:
This commit adds a Delta field to show the difference between the final profile vs the current configuration.

When the Reconcile function is called in each resource and the configurations are compared in order to check if
an update is required, the delta between those configurations is gathered and saved in the "delta" status field.

Test Plan:
PASS: Test the changes on bootstrap. Verify the new deltas are added to
    the new field and printed on logs. Verify that after the unlock, the
    deltas are empty.
PASS: On system and host resources, apply Day-2 updates. Verify the
    updates are added to the delta field. After lock/unlock, the deltas are
    empty once the resources are reconciled again.
PASS: For the other resources, apply a new configuration on Day2 and
    verify the delta is updated until the new config is applied.


enzocandotti:
Here is an explanation on how the process of capturing the delta works:

First, CollectDiffValues processes the difference between the configurations by removing data types and interfaces from each line, resulting in a set of processed lines.

Then, processLines takes these processed lines as input and iterates through them. It identifies the lines indicating changes in the configuration (prefixed with "+" or "-") and determines the corresponding parameters for these changes by invoking searchParameters. It constructs a structured report by grouping the changes under the relevant parameters, ensuring a clear representation of the differences.

Next, searchParameters analyzes the lines preceding the current line of difference. It searches for subparameters and parameters within those lines and adds them to the report. By traversing backward, it captures the context of the changes, providing a hierarchical structure to the report.

Overall, these steps enable CollectDiffValues to generate a detailed report that not only shows the specific changes in the configurations but also organizes them based on the associated parameters and subparameters, providing a comprehensive overview of the differences between the two configurations.

controllers/common/common.go
```diff
@@ -320,3 +367,133 @@ func (in *EventLogger) WarningEvent(object runtime.Object, reason string, messag
...
+				}
+			}
+		}
+
+		// Check if the line matches any parameter in the system properties.
+		for param := range parameters {
+			if strings.Contains(line, param) {
+				if i == lineNumber {
+					// If the line represents the parameter itself, indicate it with "param_found".
+					return "param_found"
```
ttakenak, line 490: Can we create constant for "param_found"?

controllers/host/host_controller.go
```diff
@@ -1383,6 +1385,42 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
 
 	logHost.Info("current config is:", "values", current)
 
+	specBytes, err := json.Marshal(profile)
```
ttakenak, line 1388: If marshal fails, can we refer the specBytes value?

controllers/host/host_controller.go
```diff
@@ -1383,6 +1385,42 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
...
+		logHost.Info("Failed to parse profile data: %s", err)
+	}
+
+	err = json.Unmarshal([]byte(currentBytes), &currentData)
+	if err != nil {
+		logHost.Info("Failed to parse current data: %s", err)
+	}
+
+	diff := cmp.Diff(currentData, specData)
+	deltaString := common.CollectDiffValues(diff, common.HostProperties)
```
ttakenak, line 1412: Not mandatory but personally, I prefer to extract these marshal/unmarshal lines to the new function.

enzocandotti: You are right, I've moved this entire block to a new function, it's clearer now. 
Thanks Tak!

controllers/host/host_controller.go
enzocandotti: Refactored. Thanks!

controllers/common/common.go
enzocandotti: Done.

controllers/datanetwork_controller.go
```diff
@@ -72,31 +78,45 @@ func dataNetworkUpdateRequired(instance *starlingxv1.DataNetwork, n *datanetwork
...
 				result = true
 			}
 		}
 	}
+	deltaString := delta.String()
+	if deltaString != "" {
+		deltaString = "\n" + strings.TrimSuffix(deltaString, "\n")
+		logDataNetwork.Info(fmt.Sprintf("delta configuration:%s\n", deltaString))
+	}
+	instance.Status.Delta = deltaString
```
ttakenak, line 115: Personally, I prefer to update status in ReconcileUpdated, so that we don't need to pass Reconciler here and easier to understand when and where status is updated. But I'd like to hear form the other reviewer.

enzocandotti: Hi Tak. Yes, I have also considered that option. The only reason I haven't implemented it that way is because in the ReconcileUpdated function, we have two return statements. So, perhaps we need to include the update status in both places.
I am open to modifying the code depending on @yjian118 's opinion.

yjian118: I'm okay with the current implementation as updating in this point explains the current config vs specs. And I believe it will get updated after the reconciliation in the next round of update_required check

yjian118 (approved):

ttakenak (approved):

PR #230, opened 2023-06-27
Implementation for add subcloud without upload

sharbhar:
Uploaded DM artifacts Helm-charts, overrides to /opt/platform


yjian118:
The commit purpose, test plan and its execution are required in  a code change

sharbhar:
Where did you get the manager_chart? Any dependencies of this commit?
[Sharang]: Will get manager_chart from 
    - set_fact:
        manager_chart: "{{ deployment_manager_chart | default('wind-river-cloud-platform-deployment-manager.tgz') }}"
The current behavior never upload the chart and overrides to the /opt/deploy/sw-version of the subcloud. Therefore, I don't think this change is necessary.
[Sharang]: Updated.

PR #231, opened 2023-06-28
Implementation of subcloud add without upload

sharbhar:
Subcloud Add without upload


sharbhar:
abandon this change, raised without testing

PR #238, opened 2023-07-25
Correct cert-manager examples

enzocandotti:
The Deployment Manager documentation examples for cert-manager is incorrect. 
We should not use the secret system-local-ca when creating issuers since it is reserved by the system.

This change updates the issuer name from system-local-ca to dm-local-ca.


yjian118 (changes requested):
The term of dm-local-ca is a little weird, can we change it to custom-local-ca instead?

enzocandotti:
Thanks guys. Will update it when we have green light on this change.

hugonicodemos (approved):

ttakenak (approved):

enzocandotti:
> Waiting for feedback from Jerry and Ayyappa before setting this PR as "ready for review".

They already approved. The change is good to go

PR #239, opened 2023-07-27
Ajust N3000 interface name change

ttakenak:
It is possible to change interface names for N3000 interface during installation. DM compares configuration before starting configure and after. As the result, insync stays false because of this name change.

This fix is to detect interface name changes and update profile if it detects.

This also includes the fix with update default values if vlan or vf interface is created during installation. Uuid for vlan/vf is created when these interfaces are created. To avoid insync=false, it updates default after these interfaces are created.

Test Plan:
PASS: Install in the environment with N3000 and
      finish successfully (reconciled=true, insync=true,
      no alarm)
PASS: Install in the environment without N3000
      and finish successfully (reconciled=true,
      insync=true, no alarm)


enzocandotti (approved):

yjian118 (changes requested):
I'm actually worrying about the deployment-config.yaml generated by deployctl. 
Can we omit the uuid in the deployment-config.yaml? Otherwise, if we generated the UUID from the unlocked system, it can not be applied the next time

yjian118 (approved):

PR #240, opened 2023-07-28
Reverting implementation of subcloud add without upload

sharbhar:
Need to set default value for user_uploaded_artifacts
if we can playbook standalone, it would fail with not-defined.


yjian118:
Why do you want to revert the previous commit? Need to add the description

yjian118 (approved):
As it is a simple revert, I'll merge it

yjian118:
Need to resolve the merge conflicts

sharbhar:
> Need to resolve the merge conflicts

No merge conflicts

PR #248, opened 2023-08-07
Add unit test

ttakenak:
- Add unit test
- Move systemconfigupdate method call to manager
- Update DefaultMaxStrategyRetryCount from test value
- Log as Error for strategy failed and timeout
- Remove duplicated logics

Test Plan:
PASS: make test (Confirm coverage is increased
      in controllers/manager)
PASS: make && DEBUG=yes make docker-build
PASS: Fresh install
PASS: Day2 operation

### DIFF
--- a/controllers/manager/manager.go
+++ b/controllers/manager/manager.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/starlingx/nfv/v1/systemconfigupdate"
 	perrors "github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


controllers/manager/manager.go
```diff
@@ -741,6 +749,26 @@ func (m *PlatformManager) GetVimClient() *gophercloud.ServiceClient {
 	return m.vimClient
 }
 
+// GcCreate is wrapper fucntion for systemconfigupdate Create
```
enzocandotti, line 752: nit: function
there are 4 occurrences

enzocandotti (approved):

PR #250, opened 2023-08-08
Correct cert-manager examples (#238)

enzocandotti:
The Deployment Manager documentation examples for cert-manager is incorrect. We should not use the secret system-local-ca when creating issuers since it is reserved by the system.

This change updates the issuer name from system-local-ca to custom-local-ca.


ttakenak (approved):

hugonicodemos (approved):
Straight cherry-pick

PR #270, opened 2023-08-28
Up version to 2.0.10-6

ttakenak:
This commit upversions to 2.0.10-6


enzocandotti (approved):

Makefile
```diff
@@ -17,7 +17,7 @@ HELM_CLIENT_VER_MAJ := $(shell echo ${HELM_CLIENT_VER} | awk -F. '{print $$2}')
 
 # Parameters for deploy tool
 GIT_HEAD := $(shell git rev-list -1 HEAD)
-GIT_LAST_TAG := $(shell git describe --abbrev=0 --tags)
+GIT_LAST_TAG := WRCP_22.12-wrs.7    # version needs to be specified here
```
yjian118, line 20: Is there other commit regarding the naming of the deployctl tool? I cannot understand why this tag can affect the name

ttakenak: deployctl tools version is obtained in cmd/deploy/cmd/versions.go from the value GIT_LAST_TAG
GIT_LAST_TAG is defined in Makefile and pass it in DEPLOY_LDFLAGS
and refereed in building deployctl
GIT_LAST_TAG is only used for deployctl.
We could hard-cord in cmd/deploy/cmd/versions.go but I prefer to implement this in Makefile because other information for deployctl version are defined in Makefile.

yjian118:
Agreed we will prefer this approach in the 2.0.10 branch, and it'll be easier to maintain the actual release this way.

yjian118 (approved):

PR #274, opened 2023-09-01
Refresh DM-monitor for DM upgrade (#249)

sharbhar:
Add capability to install/refresh DM-monitor when DM is installed/upgraded. Check if DM-monitor playbook exists and then execute then playbook to refresh DM-monitor.

Test Plan:

PASS:
- The "dm_monitor_playbook" variable is unset and the dm-monitor playbook is in default path.
- The "dm_monitor_playbook" path is provided and the dm-monitor playbook is in the given path.
- The "dm_monitor_playbook" variable is unset and the dm-monitor playbook is not in default path.
- The "dm_monitor_playbook" path is provided and the dm-monitor is not in the given path.

(cherry picked from commit 79c5bfc019ef0fa492cec9f16ee01e472d217470)


sselvara1:
LGTM

enzocandotti (approved):

PR #275, opened 2023-09-01
Adding Rescue Blocks for Timing Issues

FabrizioPerez:
In this commit, rescue blocks have been introduced after the closing and opening of the port. The aim is to identify and address various issues that may arise during these tasks.

Additionally, a potential issue related to waiting for the port to close has been resolved.

As a result, certain timers have been adjusted to enhance precision.

Test plan:
1. PASS: Deploy subcloud multiple times.
   - Verify the completion of each deployment.
   - Confirm the subcloud status is satisfactory.

2. PASS: Deploy subcloud, triggering the rescue block.
   - Ensure the execution of the rescue block.
   - Validate the continuation of the deployment workflow.

(cherry picked from commit 7864b1885844f29f44010797bbeadb7d6ae60219)

PR #277, opened 2023-09-05
Bug-fix: catch error in controller fs update (#276)

yjian118:
The error from the controller fs update in the system reconciler was overritten by the error produced when notifying the other reconcilers, therefore, if the controller fs update failed, the error will not be caught and there will be no retry afterwards.

This commit fixes this bug, avoids the error overritten and return the error in the end after notifying the dependencies.

Test plan:
Passed - deploy an AIOSX with controller_fs update, the system can be reconciled Passed - manually enforce the error response(400) in sysinv controller_fs controller, the error will be reported in the DM manager log. The controller can be unlocked, and host resource can be insync/reconciled. The retry of controller fs update can be seen in DM manager log. Passed - In case two, remove the error response, after a successful update of the controller_fs, the system resource can be insync/reconciled.

(cherry picked from commit c20c39bebecae7dd4c79169af84d79cc7751c8f5)

### DIFF
--- a/controllers/system/system_controller.go
+++ b/controllers/system/system_controller.go
@@ -1049,6 +1049,8 @@ func (r *SystemReconciler) ReconcileSystem(client *gophercloud.ServiceClient, in
 
 	err = r.ReconcileSystemFinal(client, instance, spec, info)
 	if err != nil {
+		// Need to return true to unblock the other reconcilers to work with the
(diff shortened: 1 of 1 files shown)


yjian118:
Not ready to merge, prefer wait until the merge of #275

PR #278, opened 2023-09-05
Avoid create bootstrap_finalized during reconfig (#271)

yjian118:
The bootstrap_finalized flag is only expected to be added after the bootstrap playbook, therefore, the task to add this flag is not necessary when re-applying the DM. This commit checks if the DM was installed before applying this playbook, avoids to create the flag file again if the DM was already installed.

Tests passed:
1. Applying this playbook on a freshly installed system, verified the system can be deloyed, the flag file was created.
2. Re-apply the playbook after the initial apply, verified the DM can be updated, the DM pod was re-created, the task of the flag file was skipped.

(cherry picked from commit c742ccbcfb356ff8793015cea28fe0329785c452)


ttakenak (approved):
Simple cherry-pick

PR #283, opened 2023-09-19
Ajust N3000 interface name change

ttakenak:
It is possible to change interface names for N3000 interface during installation. DM compares configuration before starting configure and after. As the result, insync stays false because of this name change.

This fix is to detect interface name changes and update profile if it detects.

This also includes the fix with update default values if vlan or vf interface is created during installation. Uuid for vlan/vf is created when these interfaces are created. To avoid insync=false, it updates default after these interfaces are created.

Update deployctl not to extract uuid values.

upversion:
charts: 2.0.10-6 -> 2.0.10-7
deployctl: wrs.9-> wrs.10

Test Plan:
PASS: Install in the environment with N3000 and
      finish successfully (reconciled=true, insync=true,
      no alarm)
PASS: Install in the environment without N3000
      and finish successfully (reconciled=true,
      insync=true, no alarm)
PASS: Extract config by deployctl and fresh install
      with the config without error

(cherry picked from commit fa4c501baabc2106863a09529fd1795f4c8f2ab1)


enzocandotti (approved):

yjian118 (changes requested):
Need to take care of the deployctl version and the helm chart version(I don't know why it was not bumped up)

ttakenak:
> Need to take care of the deployctl version and the helm chart version(I don't know why it was not bumped up)

Upversion in helm charts and deployctl.

Thank you, Yuxing!

hugonicodemos (approved):

yjian118 (approved):

PR #285, opened 2023-09-21
Check whether helmv2-cli and helmv3 are installed

ipiresso:
Starting from StarlingX 9, helmv2-cli will not be installed anymore since helmv2 has been deprecated at this point. This commit allows the DM playbook to keep working against stx mainline by checking whether helmv2-cli and helmv3 are installed, thus running only the blocks of the playbook targeted to each helm version available.

This approach is tailored to be backwards compatible in case the playbook needs to be ran against previous stx releases as well.

Test plan:
PASSED: Install stx-8 AIO-SX
                Deploy DM
	        Upgrade to stx master using a designer ISO without helmv2-cli
                Run 91-upgrade-dm.sh script during activation.
                Check if deployment-manager and dm-monitor helm releases are deployed
                and updated
                Check if DM pods are running
PASSED: Install AIO-SX master using a designer ISO without helmv2-cli
                Run DM playbook
                Check if deployment-manager and dm-monitor helm releases are deployed
                Check if DM pods are running

### DIFF
--- a/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
+++ b/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
@@ -4,6 +4,11 @@
   hosts: all
   gather_facts: false
   become: false
+
+  vars:
+    helmv2_installed: false
+    helmv3_installed: false
+
   tasks:
     - set_fact:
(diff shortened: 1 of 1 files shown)


hugonicodemos (approved):
LGTM

yjian118 (approved):

PR #288, opened 2023-09-29
Add case-insensitive support for deploymentScope

enzocandotti:
This commit extends the functionality of deploymentScope to support variations in case, including "principal", "bootstrap", "Principal", "Bootstrap", "PRINCIPAL", and "BOOTSTRAP".
The goal is to make the code more user-friendly and robust.

Test Plan:

PASS: Build image and use it for the first unlock. Verify that the host is unlocked as expected, and the scope is set to bootstrap. PASS: Perform a day-2 update using different deploymentScope values, including the six alternatives. Verify that when using uppercase, the scope is set correctly to its lowercase value.

upversion:
charts: 2.0.10-7 -> 2.0.10-8

Signed-off-by: Enzo Candotti <enzo.candotti@windriver.com>
(cherry picked from commit d357105)


enzocandotti:
@ttakenak could you confirm if we need an upversion of deployctl from wrs.10 to wrs.11? Thanks!

enzocandotti:
This commit also includes these changes: https://github.com/Wind-River/cloud-platform-deployment-manager/pull/287

ttakenak (approved):

ttakenak:
> if we need an upversion of deployctl from wrs.10 to wrs.11
I don't think we need to upversion on deployctl because this change does not affect to export configuration (status is not included in export).

yjian118 (changes requested):
Please update the version of deployctl to avoid user's confusion

yjian118 (approved):

PR #284, opened 2023-09-20
Add support for overriding the OS_REGION_NAME parameter

cmondo:
The Subcloud Name Reconfiguration feature includes a mechanism to override the OS_REGION_VALUE on the dcmanager side. This creates confusion when configuring the region value in the deployment file.

This commits adds the changes to override the OS_REGION_NAME parameter but on the DM side.

Test Plan:

PASS - Deploy a new Subcloud and check DM configuration to
       validate endpoint access. There should be no endpoint
       errors in DM logs and the subcloud should be operational.
PASS - Deploy DM to central and verify that the region has not
       been replaced, since the condition applies only to the
       subcloud role

PR #290, opened 2023-10-03
Add support for overriding the OS_REGION_NAME parameter (#284)

cmondo:
The Subcloud Name Reconfiguration feature includes a mechanism to override the OS_REGION_VALUE on the dcmanager side. This creates confusion when configuring the region value in the deployment file.

This commits adds the changes to override the OS_REGION_NAME parameter but on the DM side.

Test Plan:

PASS - Deploy a new Subcloud and check DM configuration to
       validate endpoint access. There should be no endpoint
       errors in DM logs and the subcloud should be operational.
PASS - Deploy DM to central and verify that the region has not
       been replaced, since the condition applies only to the
       subcloud role

((cherry picked from commit: 0382932693e6fa72a067bf143614921855d7aba8)

### DIFF
--- a/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
+++ b/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
@@ -349,6 +349,24 @@
             deploy_config: "/home/{{ ansible_ssh_user }}/deployment-config.yaml"
           when: inventory_hostname != 'localhost'
 
+         # Configure required region value for subcloud
+        - name: Get subcloud region value from environment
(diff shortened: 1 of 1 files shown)


hugonicodemos (approved):

yjian118:
@cmondo  Let me know when it is ready to merge, thanks

cmondo:
> @cmondo Let me know when it is ready to merge, thanks

@yjian118 - sure. now we are testing with existing subcloud. I'll let you know.

yjian118 (approved):

PR #297, opened 2023-11-02
Bug-fix: delete correct DM pod when reconfig

yjian118:
A bug was introduced in commit: #613fb38 when trying to search for
the old DM pod. This commit moves the task to search for the old
DM pod before the installation to ensure we have a unique DM pod
before installation, and deletes that pod after a reinstall.

Test plan:
Passed - deploy AIOSX subcloud
Passed - reconfig a subcloud successfully, check the dm pod name in
debug log which is not the new one in the system.

This commit also includes the support of the removal of helmv2 after
stx.9.


yjian118:
Abadon this PR to address race condition

PR #299, opened 2023-11-02
Bug-fix: Delete correct DM pod when reconfig

yjian118:
Bug-fix: Delete correct DM pod when reconfig
A bug was introduced in commit: #613fb38 when trying to search for
the old DM pod. This commit moves the task to search for the old
DM pod before the installation to ensure we have a unique DM pod
before installation, and deletes that pod after a reinstall.

Test plan:
1. Passed - successfully deploy 10 aiosx with correct image overrides
2. Passed - reconfig a subcloud successfully, check the dm pod name in
debug log which is not the new pod in the system.
3. Passed - fail the deployment of an AIOSX with wrong image tag in
overrides, then use the reconfig to correct it. Reconfig succeeded,
debug log which is not the new pod in the system.
4. Passed - repeat case 3 with 10 subclouds in batch

### DIFF
--- a/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
+++ b/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
@@ -291,6 +291,18 @@
 
       when: helmv3_installed and helmv3_dm_exists.rc == 0
 
+    - block:
+      - name: Search for the pod of the Deployment Manager
+        shell: |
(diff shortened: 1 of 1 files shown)


yjian118:
PR tested and ready to merge

enzocandotti (approved):

ttakenak (approved):

PR #305, opened 2023-12-07
Update OAM Network using oam API

hugonicodemos:
This commits enable the OAM network using the
oam network API instead of the addresspool API.

Test Plan:
1. After system installation setup the platformNetwork with new values of the OAM Network.
2. Apply the deployment configuration
3. Check the OAM Network was updated
- system oam-show
- system addrpool-list

### DIFF
--- a/controllers/platformnetwork_controller.go
+++ b/controllers/platformnetwork_controller.go
@@ -13,6 +13,7 @@ import (
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/addresspools"
 	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/networks"
+	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/oamNetworks"
 	perrors "github.com/pkg/errors"
 	starlingxv1 "github.com/wind-river/cloud-platform-deployment-manager/api/v1"
 	utils "github.com/wind-river/cloud-platform-deployment-manager/common"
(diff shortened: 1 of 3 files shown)


yjian118:
The code lgtm. But there's a request: if the deployScope = "bootstrap", ignore the difference even if the OAM network in the deployment-config.yaml is different than the current config(should be configured by the bootstrap playbook), will it be delivered in another commit?

controllers/platformnetwork_controller.go
```diff
@@ -317,6 +414,18 @@ func (r *PlatformNetworkReconciler) ReconcileAddressPool(client *gophercloud.Ser
 	} else {
 		if pool == nil {
 			pool, err = r.ReconcileNewAddressPool(client, instance)
+		} else if instance.Spec.Type == "oam" {
```
ttakenak, line 417: Personally, I prefer to use Constant for "oam" but not mandatory

ttakenak (approved):

yjian118:
Request a test of a lab deployment before merging this code. @hugonicodemos can you add two cases:
1 deploy sm-1
2 generate deploy-config.yaml with deployctl

We can separate the scope work items in the next commit if the sanity passed

sselvara1:
> This commits enable the OAM network using the oam network API instead of the addresspool API.
> 
> Test Plan:
> 
> 1. After system installation setup the platformNetwork with new values of the OAM Network.
> 2. Apply the deployment configuration
> 3. Check the OAM Network was updated
> 
> * system oam-show
> * system addrpool-list

What is the advantage of oam network API over the addresspool API for updating OAM network ?
Test plan indicates day-2 operation, right?

sselvara1:
> The code lgtm. But there's a request: if the deployScope = "bootstrap", ignore the difference even if the OAM network in the deployment-config.yaml is different than the current config(should be configured by the bootstrap playbook), will it be delivered in another commit?

Request a test case for deployScope="bootstrap" and deployScope="principal" and also capture the status for "insync" and "reconciled" in the test case.

sselvara1:
> Request a test of a lab deployment before merging this code. @hugonicodemos can you add two cases: 1 deploy sm-1 2 generate deploy-config.yaml with deployctl
> 
> We can separate the scope work items in the next commit if the sanity passed

As a follow up, edit deploy-config.yaml with OAM NW change and apply again (day-2)

controllers/platformnetwork_controller.go
hugonicodemos: We'll handle this in a follow-up commit when we create and manage the other networks (admin, mgmt). Is that okay?

hugonicodemos:
New tests are performed:
New deployment-config with platformNetwork OAM changes - day1 operation (deploymentScope = bootstrap):
- changes not allowed
- insync = false
NAME                                          TYPE   SUBNET       PREFIX   INSYNC   RECONCILED   SCOPE
platformnetwork.starlingx.windriver.com/oam   oam    10.10.20.0   24       false    true         bootstrap

hugonicodemos:
> > This commits enable the OAM network using the oam network API instead of the addresspool API.
> > Test Plan:
> > 
> > 1. After system installation setup the platformNetwork with new values of the OAM Network.
> > 2. Apply the deployment configuration
> > 3. Check the OAM Network was updated
> > 
> > 
> > * system oam-show
> > * system addrpool-list
> 
> 1. What is the advantage of oam network API over the addresspool API for updating OAM network ?
we *cant* change the OAM network via addresspool API.
> 2. Test plan indicates day-2 operation, right?
additional test for day1 added.

hugonicodemos:
> > The code lgtm. But there's a request: if the deployScope = "bootstrap", ignore the difference even if the OAM network in the deployment-config.yaml is different than the current config(should be configured by the bootstrap playbook), will it be delivered in another commit?
> 
> Request a test case for deployScope="bootstrap" and deployScope="principal" and also capture the status for "insync" and "reconciled" in the test case.

done

controllers/platformnetwork_controller.go
```diff
@@ -200,6 +263,39 @@ func (r *PlatformNetworkReconciler) ReconcileNewAddressPool(client *gophercloud.
 // ReconcileUpdated is a method which handles reconciling an existing data
```
yjian118, line 263: The description need to be allocated at the right function

yjian118:
I did the test as follows:
1 bootstrap a subcloud w/o the --deploy-config from its oam address 10.10.10.32
2 update its deployment-config with:
```
---
apiVersion: starlingx.windriver.com/v1
kind: PlatformNetwork
metadata:
  labels:
    controller-tools.k8s.io: "1.0"
  name: oam
  namespace: deployment
spec:
  allocation:
    order: random
    ranges:
    - end: 10.10.10.254
      start: 10.10.10.1
  floatingAddress: 10.10.10.42
  gateway: 10.10.10.1
  prefix: 24
  subnet: 10.10.10.0
  type: oam
``` 
3 config the subcloud with "dcmanager subcloud deploy config --deploy-config <>"

I can see the log in the DM log as:
```
2024-01-11T19:00:42.938426301Z  INFO    controller.platformnetwork.deployment/oam       delta configuration:
        +Floating Address: 10.10.10.42

2024-01-11T19:00:42.949473508Z  INFO    controller.platformnetwork.deployment/oam       updating oam network    {"uuid": "cea4fbdb-8528-49fd-bce9-a0859fa539a5", "opts": {"oam_floating_ip":"10.10.10.42"}}

``` 
After the deployment, my OAM ssh lost connection after unlock. And the deploy playbook failed as:
```
TASK [Waiting for port to become open] *****************************************
Thursday 11 January 2024  19:02:15 +0000 (0:00:38.660)       0:02:23.018 ******
fatal: [subcloud3]: FAILED! => changed=false
  elapsed: 400
  msg: Timeout when waiting for 10.10.10.32:22
``` 

I can login to the subcloud via 10.10.10.42 after the unlock:
```
4: enp2s1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 52:54:00:2a:91:ff brd ff:ff:ff:ff:ff:ff
    inet 10.10.10.42/24 brd 10.10.10.255 scope global enp2s1
       valid_lft forever preferred_lft forever
    inet6 fe80::5054:ff:fe2a:91ff/64 scope link
       valid_lft forever preferred_lft forever
``` 

Apparently, this is a failed case from the misconfiguration of the deployment-config.yaml, but we want to prevent it if possible.
@hugonicodemos do you want to address it in this commit or I can accept a follow up commit

PR #306, opened 2023-12-18
Enable deployctl to generate PlatformNetwork CRD in deployment config file

vineela-pachchipulusu:
Enable deployctl to generate PlatformNetwork CRD in deployment config file

The PlatformNetwork resources that are present in the System are not getting 
added to the deployment config file that is generated by  the deployctl tool. 
The function buildPlatformNetworks from the file build/build.go has the network 
types array which contains only type storage as of now .So the array is 
populated with all the required network types.

Test Plan:
1) PASS: To generate the deployment config yaml file where PlatformNetwork
         resources are also shown  using deployctl tool
2) PASS: To use the generated deployment config file to the bring up the system 
         and check if PlatformNetwork resources are applied.
3) PASS:  To test  initial configuration(day1)
4) PASS: To test reconfiguration(day2)

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -562,7 +562,7 @@ func (db *DeploymentBuilder) buildPlatformNetworks(d *Deployment) error {
 	}
 
 	nets := make([]*starlingxv1.PlatformNetwork, 0)
-	always_generate_networks := []string{"storage"}
+	always_generate_networks := []string{"storage", "mgmt", "oam", "admin"}
 	sort.Strings(always_generate_networks)
 	for _, p := range pools {
 		skip := false


build/build.go
```diff
@@ -562,7 +562,7 @@ func (db *DeploymentBuilder) buildPlatformNetworks(d *Deployment) error {
 	}
 
 	nets := make([]*starlingxv1.PlatformNetwork, 0)
-	always_generate_networks := []string{"storage"}
+	always_generate_networks := []string{"storage", "mgmt"}
```
yjian118, line 565: How about the oam and admin networks?

yjian118 (changes requested):
Some comment about the commit message:
1. title too long
2. need a text to explain the problem and the way to address the problem
3. need to remove JIRA as that is internal, you can create an issue to address if you want
4. the text plan doesn't cover the problem to address
5. need to test both initial and reconfiguration(day2)
6. as that is a change of deployctl, require a test to apply the configuration generated

vineela-pachchipulusu:
> Some comment about the commit message:
> 
> 1. title too long
> 2. need a text to explain the problem and the way to address the problem
> 3. need to remove JIRA as that is internal, you can create an issue to address if you want
> 4. the text plan doesn't cover the problem to address
> 5. need to test both initial and reconfiguration(day2)
> 6. as that is a change of deployctl, require a test to apply the configuration generated

Hi Yuxing, for point no 5, to test for day-2 below was the error observed

Screenshot 2024-01-04 175102

yjian118:
It indicates the 'Kind' is missing in secrets, can you compare it with the secrets in your previous deployment-config.yaml?

build/build.go
vineela-pachchipulusu: oam and admin networks are also added.

sselvara1:
> Enable deployctl to generate PlatformNetwork CRD in deployment config file
> 
> The PlatformNetwork resources that are present in the System are not getting added to the deployment config file that is generated by the deployctl tool. The function buildPlatformNetworks from the file build/build.go has the network types array which contains only type storage as of now .So the array is populated with all the required network types.
> 
> Test Plan:
> 
> 1. PASS: To generate the deployment config yaml file where PlatformNetwork
>    resources are also shown  using deployctl tool
> 2. PASS: To use the generated deployment config file to the bring up the system
>    and check if PlatformNetwork resources are applied.
> 3. PASS:  To test  initial configuration(day1)
> 4. PASS: To test reconfiguration(day2)

Below phrasing and formatting will be preferred - 

Enable deployctl to generate PlatformNetwork CRD in deployment config file

The PlatformNetwork resources that are present in the System are not getting 
added to the deployment config file that is generated by the deployctl tool.
The function buildPlatformNetworks has the network type array which contains 
only "storage" type as of now. The array is populated with the required network
types including  "mgmt", "oam", "admin".

Test Plan:

1. PASS: Generate deployment config file using deployctl wherein PlatformNetwork
             resources are captured.
2. PASS: Use the generated deployment config file to bring up the system and 
             check if PlatformNetwork resources are applied.
3. PASS: Test initial configuration (day1)
4. PASS: Test reconfiguration (day2)

sselvara1:
Code LGTM.

hugonicodemos (approved):

sselvara1 (approved):
LGTM
PASS: Test reconfiguration (day2) - this test case can be added here - 
https://github.com/Wind-River/cloud-platform-deployment-manager/pull/315

PR #309, opened 2024-01-04
Append controller domain and get IP for AUTHURL

yjian118:
After supporting the management reconfiguration in StarlingX, the openstack keystone endpoint address may change after updating the management network.

To avoid the DM cannot access the openstack endpoints after changing the endpoint address during network reconfiguration, this commit enables functions:

1. always append controller domain into the authURLs, which can be used to get the management floating IP address after DNSMask service enabled.
2. query IP address by domains in the authURLs.

Test plan:
1. Passed - deploy AIOSX with existing deployment configuratin.
2. Passed - update management network with new management subnet in the AIOSX, verified the DM can keep communicate with the sysinv API after unlock.
3. Passed - check V2 logs, verified the failure of get one IP address of a domain will not fail the client creation, verified the wrong format in the authURL will fail the client creation as expected, verified the duplicated IP address will be removed in the result.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -107,8 +107,6 @@ consider adding your user id to the local docker group.
 sudo usermod -a -G docker ${USER}
 newgrp docker
 ```
-## Install golangci
-curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.17.1
 
 ## Install other dependencies
 sudo apt-get install make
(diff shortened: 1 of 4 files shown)


DEVELOPER.md
```diff
@@ -107,8 +107,6 @@ consider adding your user id to the local docker group.
 sudo usermod -a -G docker ${USER}
 newgrp docker
 ```
-## Install golangci
```
hugonicodemos, line 110: Why are we deleting this?

hugonicodemos (approved):
LGTM, just a minor question

DEVELOPER.md
yjian118: The download of the golangci-lint is set up in the makefile as: .PHONY: golangci-lint, and it should be able to download the suitable version of the golangci-lint which is different from the README, move it in the README as it is not required and I don't want to maintain the version of golangci-lint in two places

PR #315, opened 2024-01-15
Support for Platform Network Reconfiguration

sriram-gn:
This commit enables platform network reconfiguration which is a Day-2 operation.
The scope of this commit is limited to AIO-SX systems and oam, mgmt, admin network types.
The network reconfigurtion will be ignored if the system is of other types.

Test Cases:
1. PASS - Verify network reconfig is successful on AIO-SX systems by updating platform resource configurations with deploymentScope of "principal" for oam/mgmt/admin networks.

2. PASS - Verify network reconfig is successful on AIO-SX systems by applying new platform resource configurations with deploymentScope of "principal" for oam/mgmt/admin networks.

3. PASS - Verify network reconfig is ignored on systems other than AIO-SX despite setting deploymentScope of "principal". The values for "INSYNC" and "RECONCILED" should reflect actual reconciliation status.

4. PASS - Verify that network reconfig is ignored if deploymentScope is "bootstrap" and the values for "INSYNC" and "RECONCILED" reflect the actual reconcialiation status.


hugonicodemos:
Hi @sriram-gn , I noticed we have some nested methods. Can you modularize them into helper functions?

yjian118:
@sriram-gn it is very difficult to follow so many changes together, need to modularize the functions, and if it is possible, can you work it out step by step? For instance, update the mgmt network on AIOSX first, and make a commit, and we can follow up on the other types of system and network types in the next commit

controllers/host/host_controller.go
```diff
@@ -2000,6 +2071,78 @@ func (r *HostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (r
...
+						err := r.Client.Status().Update(context.TODO(), host_instance)
+						if err != nil {
+							logHost.Error(err, "Failed to update resource for host namespace %v", host_namespace)
+							continue
+						}
+					}
+
+					err = r.TriggerDisabledHostReconcilerAndUnlock(platformClient, host_instance)
+					if err != nil {
+						logHost.Error(err, "Host reconciliation failed - trying again after 20 seconds")
```
yjian118, line 2133: It will dump too many error in the pod's log, need to reduce the frequency or change it to info v2 log for debugging only

controllers/host/host_controller.go
```diff
@@ -1543,6 +1544,76 @@ func (r *HostReconciler) ReconcileDeletedHost(client *gophercloud.ServiceClient,
...
+		host, err = hosts.Get(client, *id).Extract()
+		if err != nil {
+			if _, ok := err.(gophercloud.ErrDefault404); !ok {
+				err = perrors.Wrapf(err, "failed to get: %s", *id)
+				return err
+			}
+
+			// The resource may have been deleted by the system or operator
+			// therefore continue and attempt to recreate it.
+			logHost.Info("resource no longer exists", "id", *id)
```
yjian118, line 1565: Add a v2 in the final commit if the resource no longer exists is expected

controllers/platformnetwork_controller.go
```diff
@@ -12,8 +12,10 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/addresspools"
+	gcHosts "github.com/gophercloud/gophercloud/starlingx/inventory/v1/hosts"
 	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/networks"
 	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/oamNetworks"
+	gcSystem "github.com/gophercloud/gophercloud/starlingx/inventory/v1/system"
```
yjian118, line 18: You can use system, hosts instead of gcHosts and gcSystem following the pattern

controllers/platformnetwork_controller.go
```diff
@@ -402,12 +419,66 @@ func (r *PlatformNetworkReconciler) FindExistingAddressPool(client *gophercloud.
 	return pool, err
 }
 
+// Some of the network types such as mgmt use fixed names for address pools.
+// It's better to use FindAddressPoolByNetworkType for all network types other than network type "other".
+func (r *PlatformNetworkReconciler) FindAddressPoolByNetworkType(client *gophercloud.ServiceClient, instance *starlingxv1.PlatformNetwork) (pool *addresspools.AddressPool, err error) {
```
yjian118, line 424: Can we combine it with "FindExistingAddressPool"? You can differentiate the return in this method based on the network type though, the two functions have too many lines of duplication

controllers/platformnetwork_controller.go
```diff
@@ -661,37 +739,180 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+	if instance.Spec.Type != networks.NetworkTypeOther {
+		pool, err = r.FindAddressPoolByNetworkType(client, instance)
+	} else {
+		pool, err = r.FindExistingAddressPool(client, instance)
+	}
+	if err != nil {
+		return false, err
+	} else if pool == nil {
+		return true, nil
+	}
```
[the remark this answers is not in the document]
sriram-gn: We are not returning pool over there, it's just the err if there is any.
Since previous condition evaluates err != nil, in the `else if pool == nil` block we can be sure that there is no err and hence we are returning nil.

controllers/platformnetwork_controller.go
sriram-gn: Simplified it by using `GetAddrPoolNameByNetworkType`, this will return "management" if network type is mgmt, otherwise it would return the instance name.
This is based on the assumption that all other network pools use same name as network name.
If the assumption is incorrect and user is allowed to use different names for networks and corresponding pools (Note: system allows us to do so) then we have to switch back to using network type to determine the corresponding network pool (as in FindAddressPoolByNetworkType), which I'll be implementing in `FindExistingAddressPool` itself.
Please let me know.

controllers/host/host_controller.go
sriram-gn: Updated accordingly.

controllers/host/host_controller.go
sriram-gn: Updated accordingly.

controllers/platformnetwork_controller.go
sriram-gn: Updated accordingly.

sriram-gn:
> @sriram-gn it is very difficult to follow so many changes together, need to modularize the functions, and if it is possible, can you work it out step by step? For instance, update the mgmt network on AIOSX first, and make a commit, and we can follow up on the other types of system and network types in the next commit

> Hi @sriram-gn , I noticed we have some nested methods. Can you modularize them into helper functions? And since this adds too much complexity I'd like to see some tests.


Hi @yjian118 and @hugonicodemos , I've updated the commit to make it more modular and current HEAD (Support for Platform Network Reconfiguration) implements network reconfiguration of AIO-SX system. For other systems I'll be creating a new commit on current HEAD.

controllers/platformnetwork_controller.go
```diff
@@ -420,6 +450,13 @@ func (r *PlatformNetworkReconciler) ReconcileAddressPool(client *gophercloud.Ser
 	} else {
 		if pool == nil {
 			pool, err = r.ReconcileNewAddressPool(client, instance)
+		} else if instance.Spec.Type == cloudManager.MgmtNetworkType || instance.Spec.Type == cloudManager.AdminNetworkType {
+			if _, ok := poolUpdateRequired(instance, pool, r); ok {
+				err = r.ReconciledDeletedAddressPool(client, instance, pool)
```
hugonicodemos, line 455: This method attempts to delete an already deleted address pool, and for this, we want to the deletion. We either need to change the name of the reconcile or create a new one for this case (also, update the existing method comment, please).

controllers/platformnetwork_controller.go
```diff
@@ -420,6 +450,13 @@ func (r *PlatformNetworkReconciler) ReconcileAddressPool(client *gophercloud.Ser
 	} else {
 		if pool == nil {
 			pool, err = r.ReconcileNewAddressPool(client, instance)
+		} else if instance.Spec.Type == cloudManager.MgmtNetworkType || instance.Spec.Type == cloudManager.AdminNetworkType {
```
hugonicodemos, line 453: Add a comment explain this user case

controllers/platformnetwork_controller.go
```diff
@@ -661,37 +698,220 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
-				err = r.ReconcileNetwork(client, instance)
+	if len(all_hosts) == 1 {
+		for _, host := range all_hosts {
+			if host.Personality == "controller" {
+				default_system, err := system.GetDefaultSystem(client)
+				if err != nil {
+					return nil, false, common.NewSystemDependency("Failed to get default system")
+				}
+
+				if default_system.SystemMode == "simplex" && default_system.SystemType == "All-in-one" {
```
hugonicodemos, line 716: why are you not using this to check? Also use consts.

controllers/platformnetwork_controller.go
```diff
@@ -661,37 +698,220 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+		if networkResourceRequired(instance) {
+			err = r.ReconcileNetwork(client, instance)
+		}
+	}
+	return err
+}
+
+// ReconcileResource interacts with the system API in order to reconcile the
+// state of a data network with the state stored in the k8s database.
+func (r *PlatformNetworkReconciler) ReconcileResource(client *gophercloud.ServiceClient, instance *starlingxv1.PlatformNetwork) (err error) {
```
hugonicodemos, line 829: I still believe this can be further modularized.

hugonicodemos (changes requested):
@sriram-gn, why isn't the networking reconcile responsible for the network changes that you're making?

controllers/platformnetwork_controller.go
```diff
@@ -385,15 +402,25 @@ func (r *PlatformNetworkReconciler) FindExistingAddressPool(client *gophercloud.
...
-			if p.Name == instance.Name {
-				pool = &p
+		results, err := networks.ListNetworks(client)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, nwk := range results {
+			if nwk.Type == instance.Spec.Type {
+				for _, p := range addr_pools {
```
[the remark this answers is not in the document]
sriram-gn: FindExistingAddressPool should be returning pool of type *addresspools.AddressPool.
From the address pool alone we will not know if it's being used by network of particular type.
Condition 1: `if nwk.Type == instance.Spec.Type` helps determine what PoolUUID I should be looking for.
Condition 2: `if p.ID == nwk.PoolUUID` helps determine what *addresspools.AddressPool I should be returning.

sriram-gn:
> @sriram-gn, why isn't the networking reconcile responsible for the network changes that you're making?

Are we talking about `ReconcileNetwork` method? If so, yes, it is responsible for making network changes.

controllers/platformnetwork_controller.go
```diff
@@ -146,7 +148,7 @@ func oamUpdateRequired(instance *starlingxv1.PlatformNetwork, p *oamNetworks.OAM
 // include in the request options to minimum churn and to ease debugging.
 func poolUpdateRequired(instance *starlingxv1.PlatformNetwork, p *addresspools.AddressPool, r *PlatformNetworkReconciler) (opts addresspools.AddressPoolOpts, result bool) {
 	var delta strings.Builder
-	if instance.Name != p.Name {
+	if r.GetAddrPoolNameByNetworkType(instance.Spec.Type, instance.Name) != p.Name {
```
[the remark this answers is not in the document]
sriram-gn: > Can we add a comment here referring to the rule about the mgmt address pool name? We need to keep tracking this since it's possible to change in the future.

Done.

controllers/platformnetwork_controller.go
```diff
@@ -385,15 +402,25 @@ func (r *PlatformNetworkReconciler) FindExistingAddressPool(client *gophercloud.
...
 			err = perrors.Wrap(err, "failed to list pools")
 			return nil, err
 		}
 
-		for _, p := range results {
-			if p.Name == instance.Name {
-				pool = &p
+		results, err := networks.ListNetworks(client)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
sriram-gn: > use err = perrors.Wrap(err, "failed to list networks")

Done.

controllers/platformnetwork_controller.go
sriram-gn: > This method attempts to delete an already deleted address pool, and for this, we want to the deletion. We either need to change the name of the reconcile or create a new one for this case (also, update the existing method comment, please).

Done.

controllers/platformnetwork_controller.go
sriram-gn: > Add a comment explain this user case

Done.

controllers/platformnetwork_controller.go
sriram-gn: > I still believe this can be further modularized.

Done.

controllers/platformnetwork_controller.go
sriram-gn: Considering SX -> DX upgrade or other such scenarios leading to propagation of stale entries we are using two other conditions `if len(all_hosts) == 1` and `if host.Personality == cloudManager.PersonalityController` to be on the safer side.
Updated to use consts as suggested but keeping the conditions for now, if you still think we can manage it just the `default_system.SystemMode` and `default_system.SystemType` do let me know, I'll remove it.

yjian118:
All the checks are failed, please make sure all the integrated checks are passed

controllers/manager/manager.go
```diff
@@ -444,6 +464,38 @@ func (m *PlatformManager) GetPlatformClient(namespace string) *gophercloud.Servi
 	return nil
 }
 
+func (m *PlatformManager) SendHostStrategyUpdate(host_strategy HostStrategyInfo) error {
```
yjian118, line 467: Need to add description on every new methods, difficult to read & understand

controllers/manager/manager.go
```diff
@@ -444,6 +464,38 @@ func (m *PlatformManager) GetPlatformClient(namespace string) *gophercloud.Servi
...
+	return host_strategy
+}
+
+func (m *PlatformManager) SendHostReconciliationTrigger(host_strategy HostStrategyInfo) error {
+	m.HostReconciliationChannel <- host_strategy
+	return nil
+}
+
+func (m *PlatformManager) ReceiveHostReconciliationTrigger() HostStrategyInfo {
+	host_strategy := <-m.HostReconciliationChannel
```
yjian118, line 483: Please avoid to use "channel" in the name, that is obviously a channel, no need to mention, the naming convention should be verb+object

controllers/host/host_controller.go
```diff
@@ -2073,6 +2147,78 @@ func (r *HostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (r
 		return common.RetrySystemNotReady, nil
 	}
 
+	if !r.CloudManager.GetHostUpdateRoutinesRunning() {
+		logHost.Info("Starting goroutines to receive host strategy updates from other controllers")
```
yjian118, line 2151: This log should not be printed to the user

controllers/host/host_controller.go
```diff
@@ -2073,6 +2147,78 @@ func (r *HostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (r
...
+		go func() {
```
yjian118, line 2152: The two go-routines should not be looping infinitely except for a grace shutdown.

controllers/manager/manager.go
```diff
@@ -444,6 +464,38 @@ func (m *PlatformManager) GetPlatformClient(namespace string) *gophercloud.Servi
...
+func (m *PlatformManager) GetHostUpdateRoutinesRunning() bool {
+	m.lock.Lock()
+	defer func() { m.lock.Unlock() }()
+	return m.IsHostUpdateRoutineRunning
+}
+
+func (m *PlatformManager) SetHostUpdateRoutinesRunning(b bool) {
+	m.lock.Lock()
+	defer func() { m.lock.Unlock() }()
+	m.IsHostUpdateRoutineRunning = true
```
yjian118, line 496: The parameter b here is not used, and I don't find a place to set it to false

controllers/host/host_controller.go
```diff
@@ -1616,6 +1617,79 @@ func (r *HostReconciler) ReconcileDeletedHost(client *gophercloud.ServiceClient,
 	return nil
 }
 
+// TriggerNetworkReconcilerAndUnlock is used to trigger network reconciliation after making changes to platform network.
+// After successful reconciliation unlock-strategy is applied if the system is in locked state.
+func (r *HostReconciler) TriggerNetworkReconcilerAndUnlock(client *gophercloud.ServiceClient, instance *starlingxv1.Host) (err error) {
```
hugonicodemos, line 1622: The host_controller already do a reconcile when the host is locked disabled. Is there any special case that we need to trigger this again?

controllers/host/host_controller.go
hugonicodemos, line 2152: the goroutines could cause issues in the long term, we are keeping track of leaks here?

controllers/host/host_controller.go
```diff
@@ -2073,6 +2147,78 @@ func (r *HostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (r
...
+					if err != nil {
+						logHost.Error(err, "Failed to update status %v", host_instance.Status)
+						continue
+					}
+				}
+
+				if !r.CloudManager.GetStrageySent() {
+					msg := fmt.Sprintf("Applying %s strategy on host %s", host_strategy.StrategyRequired, host_instance.Name)
+					logHost.Info(msg)
+					r.CloudManager.SetResourceInfo(cloudManager.ResourceHost, host_strategy.Host.Personality, host_instance.Name, host_instance.Status.Reconciled, host_instance.Status.StrategyRequired)
```
hugonicodemos, line 2175: create variable to keep this line shorter. At least for host_instance.Status.Reconciled and host_instance.Status.StrategyRequired

controllers/host/host_controller.go
```diff
@@ -2073,6 +2147,78 @@ func (r *HostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (r
...
+					m := NewLockedDisabledHostMonitor(host_instance, host_strategy.Host.ID)
+					_ = r.CloudManager.StartMonitor(m, msg)
+				} else {
+					logHost.Info("Ignoring host strategy update, already sent.")
+				}
+
+			}
+		}()
+
+		go func() {
```
hugonicodemos, line 2185: ditto

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +741,261 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+						if err != nil {
+							return err
+						}
+					}
+
+					return nil
+				}
 			}
+
+			// }
```
hugonicodemos, line 998: delete

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +741,261 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+				return err2
+			} else {
+				// Try to reconcile rest of the networks and update the status accordingly.
+				_, err = r.ReconcileAddressPoolAndNetwork(client, instance)
+				if err != nil {
+					return err
+				}
+			}
+		} else if instance.Status.DeploymentScope == cloudManager.ScopePrincipal {
+			// if update_pn {
```
hugonicodemos, line 910: delete

hugonicodemos (changes requested):

controllers/platformnetwork_controller.go
```diff
@@ -335,9 +388,9 @@ func (r *PlatformNetworkReconciler) ReconcileUpdatedAddressPool(client *gophercl
 	return nil
 }
 
-// ReconcileNew is a method which handles reconciling a new data resource and
-// creates the corresponding system resource thru the system API.
-func (r *PlatformNetworkReconciler) ReconciledDeletedAddressPool(client *gophercloud.ServiceClient, instance *starlingxv1.PlatformNetwork, pool *addresspools.AddressPool) error {
```
sselvara1, line 340: what is the purpose of renaming this API? can we not still use the same API name and have the functional changes ..

controllers/platformnetwork_controller.go
```diff
@@ -407,19 +471,31 @@ func (r *PlatformNetworkReconciler) FindExistingAddressPool(client *gophercloud.
 // the network resource since at the system level these are two independent
 // resources.
 func (r *PlatformNetworkReconciler) ReconcileAddressPool(client *gophercloud.ServiceClient, instance *starlingxv1.PlatformNetwork) error {
-	pool, err := r.FindExistingAddressPool(client, instance)
+	var pool *addresspools.AddressPool
```
sselvara1, line 474: is there a specific reason to explicitly call out the data type - *addresspools.AddressPool ?

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +741,261 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
 	return !instance.Status.DeepEqual(status)
 }
 
-// ReconcileResource interacts with the system API in order to reconcile the
-// state of a data network with the state stored in the k8s database.
-func (r *PlatformNetworkReconciler) ReconcileResource(client *gophercloud.ServiceClient, instance *starlingxv1.PlatformNetwork) (err error) {
```
sselvara1, line 666: the diff is misleading if ReconcileResource API is removed.

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +741,261 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+	return false, nil
+}
+
+func (r *PlatformNetworkReconciler) NetworkUpdateRequired(client *gophercloud.ServiceClient, instance *starlingxv1.PlatformNetwork) (bool, error) {
+	if instance.Spec.Type == cloudManager.OAMNetworkType {
+		oamNetworkList, err := oamNetworks.ListNetworks(client)
+		if err != nil {
+			return false, err
+		} else if len(oamNetworkList) == 0 {
+			return true, nil
```
sselvara1, line 793: what does 'true' indicate here? update required?

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +741,261 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+			if _, ok := oamUpdateRequired(instance, oamNetwork, r); ok {
+				return true, nil
+			}
+		}
+	} else {
+		network, err := r.FindExistingNetwork(client, instance)
+		if err != nil {
+			return false, err
+		} else if network == nil {
+			return true, nil
```
sselvara1, line 805: what does 'true' indicate here? update required? what will be updated?

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +741,261 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+		err2 := r.Client.Status().Update(context.TODO(), instance)
+		if err2 != nil {
+			logPlatformNetwork.Error(err2, "failed to update platform network status")
+			return inSync, err2
+		}
+	}
+	return inSync, nil
+}
 
-		if !instance.Status.Reconciled {
```
sselvara1, line 685: where is this code getting removed from ? either of the above and below functions..

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +741,261 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+			StrategyRequired: strategy,
+			Host:             *host,
 		}
+		_ = r.CloudManager.SendHostReconciliationTrigger(host_strategy)
+		return true
+	}
+	return false
+}
 
-		if instance.Status.Reconciled {
```
sselvara1, line 690: where is this code getting removed from ? either of the above and below functions..

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +741,261 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+	inSync, err2 := r.UpdateInsyncStatus(client, instance, err)
+	if err2 != nil {
+		return inSync, err2
+	}
+
+	return inSync, err
+}
+
+// ReconcileResource interacts with the system API in order to reconcile the
+// state of a data network with the state stored in the k8s database.
```
sselvara1, line 879: platform network

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +741,261 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+	oldStatus := instance.Status.DeepCopy()
+
+	if instance.DeletionTimestamp.IsZero() {
+		update_pn, err := r.PlatformNetworkUpdateRequired(client, instance)
+		if err != nil {
+			return err
+		}
+		if instance.Status.DeploymentScope == cloudManager.ScopeBootstrap {
+			inSync := !update_pn
+			if instance.Spec.Type == cloudManager.MgmtNetworkType || instance.Spec.Type == cloudManager.OAMNetworkType || instance.Spec.Type == cloudManager.AdminNetworkType {
```
sselvara1, line 891: If a line feels too long, it should be refactored instead of broken.

controllers/platformnetwork_controller.go
```diff
@@ -851,7 +1158,6 @@ func (r *PlatformNetworkReconciler) UpdateConfigStatus(instance *starlingxv1.Pla
 					instance.Status.Reconciled = false
 					// Update strategy required status for strategy monitor
 					r.CloudManager.UpdateConfigVersion()
-					r.CloudManager.SetResourceInfo(cloudManager.ResourcePlatformnetwork, "", instance.Name, instance.Status.Reconciled, cloudManager.StrategyNotRequired)
```
sselvara1, line 854: is this part of ReconcileResource?

controllers/manager/manager.go
sriram-gn: > Need to add description on every new methods, difficult to read & understand

Done.

controllers/manager/manager.go
sriram-gn: > Please avoid to use "channel" in the name, that is obviously a channel, no need to mention, the naming convention should be verb+object

Done.

controllers/platformnetwork_controller.go
```diff
@@ -12,8 +12,10 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/addresspools"
+	hosts "github.com/gophercloud/gophercloud/starlingx/inventory/v1/hosts"
```
sselvara1, line 15: Need to load diff - large diffs are not rendered by default.
do we have too many changes to accommodate in this PR?

go.mod
```diff
@@ -92,4 +92,4 @@ require (
 )
 
 // replace github.com/gophercloud/gophercloud => ./external/gophercloud
-replace github.com/gophercloud/gophercloud => github.com/Wind-River/gophercloud v0.0.0-20231207152242-d06bd3ca4f52
```
sselvara1, line 95: I think we need to make sure this gets updated in final version of the merge after gopher cloud commit is merged.

helm/wind-river-cloud-platform-deployment-manager/templates/crds.yaml
```diff
@@ -1907,6 +1907,12 @@ spec:
                 required:
                 - type
                 type: object
+              controller0Address:
```
sselvara1, line 1910: is this for duplex setup? i see scope is for simplex setup, how does C-0 and C-1 impact ?

controllers/host/host_controller.go
sriram-gn: Implemented a method to exit out of loop / goroutine in manager.go - `CloseHostRoutine()`.

controllers/host/host_controller.go
```diff
@@ -1616,6 +1617,79 @@ func (r *HostReconciler) ReconcileDeletedHost(client *gophercloud.ServiceClient,
...
+	if err != nil {
+		return err
+	}
+
+	err = r.ReconcileNetworking(client, instance, profile, &hostInfo)
+	if err != nil {
+		return err
+	}
+
+	if host.IsLockedDisabled() {
```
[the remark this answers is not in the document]
sriram-gn: We are keeping it inside the function because `IsLockedDisabled` is part of gophercloud hosts and we are already making an API call inside this method. 
Having it outside the function would mean repeating the API call outside as well (more lines of code and an API call).
Works fine for now, I think we can move it outside if the use case for this expands in future.

controllers/host/host_controller.go
sriram-gn: In order to trigger reconciliation from host controller and use VIM strategy to unlock, we have to set the deploymentScope to `principal` - I think it's not good idea to update deploymentScope without getting it from the user.

Just updating `instance.Status.Reconciled` to false and letting the `HostReconciler` handle reconciliation would mean, the `HostReconciler` would use Day-1 procedure, ie. direct API call to sysinv and not use VIM strategy to unlock the system.

Given the use case, we are fixing interface-network association from outisde of `Reconcile` body.

controllers/host/host_controller.go
sriram-gn: Refactored the code to have a better lifecycle management of this goroutine, please take a look.
Although it looks like we are infinitely looping around, the code in the loop is executed only when data is sent through corresponding channel and once the code is executed the loop is blocked waiting to receive data in upcoming iterations.
Since it's considered ok to generally keep the channel open all the time, I think it's safe to that extent.

controllers/host/host_controller.go
sriram-gn: > create variable to keep this line shorter. At least for host_instance.Status.Reconciled and host_instance.Status.StrategyRequired

Done

controllers/platformnetwork_controller.go
sriram-gn: > delete

Done.

controllers/platformnetwork_controller.go
sriram-gn: > delete

Done.

controllers/host/host_controller.go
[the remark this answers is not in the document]
sriram-gn: > please add comment specifying the purpose of this go routine.

Done.

controllers/host/host_controller.go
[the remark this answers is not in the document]
sriram-gn: > please add comment specifying the purpose of this go routine.

This goroutine has been removed and functionality is being handled in above goroutine itself.

controllers/manager/manager.go
```diff
@@ -49,7 +50,9 @@ const (
 
 const (
 	OAMNetworkType   = "oam"
+	MgmtNetworkType  = "mgmt"
 	AdminNetworkType = "admin"
+	MgmtAddrPoolName = "management"
```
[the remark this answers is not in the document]
sriram-gn: This is expected, for platform network of type "mgmt" sysinv will reject any names other than the name "management" for the address pool. We have also mentioned this dependency in a comment in `platformnetwork_controller.go`

controllers/manager/manager.go
```diff
@@ -171,13 +180,21 @@ type StrategyStatus struct {
 	MonitorStarted bool
 }
 
+type HostStrategyInfo struct {
+	StrategyRequired string
```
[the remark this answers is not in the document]
sriram-gn: The values of StrategyRequired can be one of `(not_required, lock_required, unlock_required)`.
It should be of type string as we have it currently.

controllers/platformnetwork_controller.go
sriram-gn: > If a line feels too long, it should be refactored instead of broken.

Done.

go.mod
sriram-gn: > I think we need to make sure this gets updated in final version of the merge after gopher cloud commit is merged.

Yes, gophercloud commit is already merged and the new gophercloud version reference is already part of this PR.

helm/wind-river-cloud-platform-deployment-manager/templates/crds.yaml
sriram-gn: https://docs.starlingx.io/dist_cloud/kubernetes/update-a-subcloud-network-parameters-b76377641da4.html#aio-sx-subcloud

Under Step 2 of "AIO-SX subcloud" we have:

`~(keystone_admin)]$ system addrpool-add --floating-address 192.168.103.2 --controller0-address 192.168.103.3 --controller1-address 192.168.103.4 --gateway-address 192.168.103.1 admin 192.168.103.0 24`

The procedure says we need to provide all of floating address, c0 and c1 IPs, hence we have incorporated it accordingly.

controllers/host/host_controller.go
yjian118: Where does it being called?

yjian118: Further, request a change of the methods name with "Routine", please use the purpose of the methods instead

yjian118: Request a end of the goroutine threads when the main thread closed

sselvara1:
Also, please below test case for https://github.com/Wind-River/cloud-platform-deployment-manager/pull/306
PASS: Test reconfiguration (day2) - this test case can be added here -

controllers/host/host_controller.go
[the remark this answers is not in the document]
hugonicodemos: I mean, the method should only "DO" one thing, meaning this should be responsible to only trigger the network reconcile and we do the unlock after the method call.

controllers/platformnetwork_controller.go
hugonicodemos: you're renaming the import using the same name. just do:
"github.com/gophercloud/gophercloud/starlingx/inventory/v1/hosts" without hosts

controllers/platformnetwork_controller.go
```diff
@@ -12,8 +12,10 @@ import (
...
 	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/networks"
 	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/oamNetworks"
+	system "github.com/gophercloud/gophercloud/starlingx/inventory/v1/system"
```
hugonicodemos, line 18: ditto

controllers/platformnetwork_controller.go
```diff
@@ -407,19 +471,31 @@ func (r *PlatformNetworkReconciler) FindExistingAddressPool(client *gophercloud.
 // the network resource since at the system level these are two independent
 // resources.
 func (r *PlatformNetworkReconciler) ReconcileAddressPool(client *gophercloud.ServiceClient, instance *starlingxv1.PlatformNetwork) error {
-	pool, err := r.FindExistingAddressPool(client, instance)
+	var pool *addresspools.AddressPool
+	var err error
+	pool, err = r.FindExistingAddressPool(client, instance)
```
hugonicodemos, line 476: you can declare and assign the variables using `:=` (check the line 410 removed)

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +741,267 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
 		}
+	}
+
+	return nil, false, nil
+}
+
+func (r *PlatformNetworkReconciler) AddressPoolUpdateRequired(client *gophercloud.ServiceClient, instance *starlingxv1.PlatformNetwork) (bool, error) {
+	var pool *addresspools.AddressPool
+	var err error
+	pool, err = r.FindExistingAddressPool(client, instance)
```
hugonicodemos, line 774: ditto about declare and assign

controllers/host/host_controller.go
```diff
@@ -1636,7 +1716,7 @@ func (r *HostReconciler) ReconcileResource(client *gophercloud.ServiceClient, in
 
 			// The resource may have been deleted by the system or operator
 			// therefore continue and attempt to recreate it.
-			logHost.Info("resource no longer exists", "id", *id)
+			logHost.V(2).Info("resource no longer exists", "id", *id)
```
yjian118, line 1719: Suggest to leave this log level un-changed only if it produce too many logs

controllers/manager/manager.go
```diff
@@ -38,8 +39,9 @@ const SystemEndpointSecretName = "system-endpoint"
 
 const (
 	// Defines annotation keys for resources.
-	NotificationCountKey = "deployment-manager/notifications"
-	ReconcileAfterInSync = "deployment-manager/reconcile-after-insync"
+	NotificationCountKey    = "deployment-manager/notifications"
+	ReconcileHostNetworking = "deployment-manager/reconcile-host-networking"
```
yjian118, line 43: ReconcileHostByPlatformNetwork

controllers/manager/manager.go
```diff
@@ -421,7 +435,11 @@ func (m *PlatformManager) NotifySystemDependencies(namespace string) error {
 }
 
 func (m *PlatformManager) NotifyResource(object client.Object) error {
-	return m.notifyController(object)
+	return m.notifyController(object, NotificationCountKey, false)
+}
+
+func (m *PlatformManager) NotifyHostReconciliation(object client.Object, deleteKey bool) error {
```
yjian118, line 441: NotifyHostController

controllers/manager/manager.go
```diff
@@ -49,7 +51,9 @@ const (
 
 const (
 	OAMNetworkType   = "oam"
+	MgmtNetworkType  = "mgmt"
```
yjian118, line 54: I would prefer to assign them to the reconciler/controllers, it looks messy in the following switch conditions, can you add a TODO if you are not going to re-position them?

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +742,288 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+			}
 		}
+	} else {
+		network, err := r.FindExistingNetwork(client, instance)
+		if err != nil {
+			return false, err
+		} else if network == nil {
+			return true, nil
+		}
+		if _, ok := plNetworkUpdateRequired(instance, network); ok {
```
yjian118, line 808: Now we have plNetworkUpdateRequired and PlatformNetworkUpdateRequried which is very confusing, can we add a comment on the new method do explain them, and better name them?

controllers/manager/manager.go
```diff
@@ -464,36 +464,54 @@ func (m *PlatformManager) GetPlatformClient(namespace string) *gophercloud.Servi
...
 	return m.IsHostUpdateRoutineRunning
 }
 
+// SetHostUpdateRoutinesRunning is used to set IsHostUpdateRoutineRunning
+// depending upon the status of the go routines.
 func (m *PlatformManager) SetHostUpdateRoutinesRunning(b bool) {
 	m.lock.Lock()
 	defer func() { m.lock.Unlock() }()
-	m.IsHostUpdateRoutineRunning = true
+	m.IsHostUpdateRoutineRunning = b
```
sselvara1: can we have meaningful name instead of "b"

controllers/manager/monitor_test.go
```diff
@@ -369,17 +369,23 @@ func (m *Dummymanager) GetHostUpdateRoutinesRunning() bool {
...
 }
-func (m *Dummymanager) SendHostReconciliationTrigger(host_strategy HostStrategyInfo) error {
-	return nil
+func (m *Dummymanager) CloseHostRoutine() {
+
+}
+func (m *Dummymanager) GetHostReconciliationStatus() bool {
+	return false
+}
+func (m *Dummymanager) SetHostReconciliationStatus(b bool) {
```
sselvara1: what is purpose of these empty functions ?

examples/aio-sx/vxlan/vxlan_platform_networks.yaml
```diff
@@ -73,3 +73,22 @@ spec:
     ranges:
       - start: fd00:0:0:2::2
         end: fd00:0:0:2::a
+---
```
sselvara1, line 76: is vxlan the proper path?

sselvara1 (commented):
LGTM. Please make sure testing is covered.

controllers/manager/monitor_test.go
sriram-gn: We've removed all these functions.

controllers/manager/manager.go
sriram-gn: We've removed this function.

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +742,291 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
 	return !instance.Status.DeepEqual(status)
 }
 
-// ReconcileResource interacts with the system API in order to reconcile the
-// state of a data network with the state stored in the k8s database.
-func (r *PlatformNetworkReconciler) ReconcileResource(client *gophercloud.ServiceClient, instance *starlingxv1.PlatformNetwork) (err error) {
-	oldStatus := instance.Status.DeepCopy()
+// Determine if the system type is All-in-one simplex, return the Host object if so.
+func (r *PlatformNetworkReconciler) AIOSXHost(client *gophercloud.ServiceClient) (*hosts.Host, bool, error) {
```
yjian118, line 746: This method can be simplified as: GetDefaultSystem -> default_system.SystemMode == SystemModeSimplex -> true

The count of the hosts and the check of the system type is not necessary

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +742,291 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+		r.CloudManager.SetResourceInfo(cloudManager.ResourceHost, host_strategy.Host.Personality, host_instance.Name, reconciled, strategy)
+		m := hostController.NewLockedDisabledHostMonitor(host_instance, host_strategy.Host.ID)
+		_ = r.CloudManager.StartMonitor(m, msg)
+	}
+	return common.NewResourceConfigurationDependency("waiting for system to be in locked state before reconfiguring network")
+}
+
+// ReconcileResource interacts with the system API in order to reconcile the
+// state of a data network with the state stored in the k8s database.
+func (r *PlatformNetworkReconciler) ReconcileResource(client *gophercloud.ServiceClient, instance *starlingxv1.PlatformNetwork, request_namespace string) (err error) {
```
yjian118, line 913: Expect to break this long method by scenarios: ReconcileResourceBootstrap, ReconcileResourcePincipal(ReconcileResourcePincipalSimplex, ReconcileResourcePincipalMultinodes)

controllers/platformnetwork_controller.go
```diff
@@ -661,38 +742,291 @@ func (r *PlatformNetworkReconciler) statusUpdateRequired(instance *starlingxv1.P
...
+								if err != nil {
+									return err
+								}
+							}
+
+							// Trigger network reconciliation on host controller by creating annotation
+							// on host instance if the network type is management.
+							host_instance, err := r.GetHostInstance(host.Hostname, request_namespace)
+							if err != nil {
+								logPlatformNetwork.Error(err, "Failed to get resource for host namespace %v", request_namespace)
```
yjian118, line 971: Failed to get host resource from namespace, similiar comments apply in the rest logs

controllers/platformnetwork_controller.go
sriram-gn: > This method can be simplified as: GetDefaultSystem -> default_system.SystemMode == SystemModeSimplex -> true
> 
> The count of the hosts and the check of the system type is not necessary

Done.

controllers/platformnetwork_controller.go
sriram-gn: > Expect to break this long method by scenarios: ReconcileResourceBootstrap, ReconcileResourcePincipal(ReconcileResourcePincipalSimplex, ReconcileResourcePincipalMultinodes)

Done.

controllers/platformnetwork_controller.go
sriram-gn: > Failed to get host resource from namespace, similiar comments apply in the rest logs

Done.

controllers/platformnetwork_controller.go
```diff
@@ -829,35 +824,38 @@ func (r *PlatformNetworkReconciler) IsPlatformNetworkInSync(client *gophercloud.
 		return false, err
 	}
 
-	return pool_update_required || nwk_update_required, err
+	return !(pool_update_required || nwk_update_required), err
```
sselvara1, line 827: what's the change now?

controllers/platformnetwork_controller.go
```diff
@@ -883,7 +881,6 @@ func (r *PlatformNetworkReconciler) GetHostInstance(hostname string, request_nam
 func (r *PlatformNetworkReconciler) HandleHostStrategyUpdates(host_strategy cloudManager.HostStrategyInfo, request_namespace string) (err error) {
 	host_instance, err := r.GetHostInstance(host_strategy.Host.Hostname, request_namespace)
 	if err != nil {
-		logPlatformNetwork.Error(err, "Failed to get resource for host namespace %v", request_namespace)
```
sselvara1: is the log message removed?

sselvara1 (commented):
Looks much better.
Please confirm the testing is covered. I can approve.

controllers/platformnetwork_controller.go
sriram-gn: Hugo suggested to change the name since it's deleting the address to reconcile and not reconciling "deleted" address pool.

controllers/platformnetwork_controller.go
sriram-gn: Previously there was an implementation to `FindAddressPoolByNetworkType` which is merged with `FindExistingAddressPool` currently.
There was if and else statement used at the very beginning to obtain the pool from these methods depending upon network type, initializing the vars inside if-else block would mean scope of the var is limited to only that block which wouldn't work.

controllers/platformnetwork_controller.go
sriram-gn: It's not removed, it's refactored / updated.

controllers/platformnetwork_controller.go
sriram-gn: Yes.

controllers/platformnetwork_controller.go
sriram-gn: Returning `true` indicates applied config doesn't match with running config and update is required.

controllers/platformnetwork_controller.go
sriram-gn: Done.

controllers/platformnetwork_controller.go
sriram-gn: Suggest to compare with original file for clarity, these diffs might be misleading due to too many changes.

controllers/platformnetwork_controller.go
sriram-gn: Same as above.

controllers/platformnetwork_controller.go
sriram-gn: Previously the method was called `UpdatePlatformNetwork` and Yuxing suggested to change the name and it was renamed `IsPlatformNetworkInSync`, updated return value to match the name and method description.

controllers/platformnetwork_controller.go
sriram-gn: yes, I noticed we are already logging it on `GetHostInstance`, hence removed it from other places.

hugonicodemos (approved):
I'll let @yjian118 merge if he has no more concerns.

yjian118 (approved):

PR #319, opened 2024-01-23
Check os region name before update it

yjian118:
In some cases, the deployment config doesn't contain the os region name, the ansible lineinfile will append the os region name to the end of the configuration file which result failed to apply to the cluster.

This commit adds pre-checks of the os region name to skip the region name modification to prevent this issue.

Test plan:
1. Deploy AIOSX stand alone, and update it with the playbook.
2. Deploy AIOSX subcloud, and update it by 'dcmanager subcloud deploy config' with empty model.
3. Deploy AIOSX subcloud, and update it by 'dcmanager subcloud deploy config' with full model.

### DIFF
--- a/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
+++ b/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
@@ -368,22 +368,20 @@
             deploy_config: "/home/{{ ansible_ssh_user }}/deployment-config.yaml"
           when: inventory_hostname != 'localhost'
 
-         # Configure required region value for subcloud
-        - name: Get subcloud region value from environment
-          shell: source /etc/platform/openrc; echo $OS_REGION_NAME
-          register: os_region_name_from_env
+        - name: Configure required region value for subcloud
+          block:
+          - name: Get subcloud region value from environment
+            shell: source /etc/platform/openrc; echo $OS_REGION_NAME
+            register: os_region_name_from_env
 
-        - name: Show subcloud region
-          debug:
-            msg:
-            - "Subcloud region value : {{os_region_name_from_env.stdout}}"
+          - name: Show subcloud region
+            debug:
+              msg:
+              - "Subcloud region value : {{os_region_name_from_env.stdout}}"
 
-        - name: Configure new subcloud region value
-          lineinfile:
-            path: "{{deploy_config}}"
-            regexp: 'OS_REGION_NAME'
-            line: "  OS_REGION_NAME: {{os_region_name_from_env.stdout}}"
-            state: present
+          - name: Configure new subcloud region value
+            command: |
+              sed -i 's/^\(\s*OS_REGION_NAME:\).*/\1 {{ os_region_name_from_env.stdout }}/' {{ deploy_config }}
           when: ("subcloud" in get_distributed_cloud_role.stdout)
 
         - wait_for:


sriram-gn:
Hi @yjian118 ,

I think by setting backrefs to true, we can avoid using `grep`.

Something like following snippet:
```
- name: Configure new subcloud region value
            lineinfile:
              path: "{{deploy_config}}"
              regexp: 'OS_REGION_NAME'
              line: "  OS_REGION_NAME: {{os_region_name_from_env.stdout}}"
              backrefs: true
              state: present
          when: ("subcloud" in get_distributed_cloud_role.stdout) and os_region_name_exists.rc == 0
```

sselvara1:
> Hi @yjian118 ,
> 
> I think by setting backrefs to true, we can avoid using `grep`.
> 
> Something like following snippet:
> 
> ```
> - name: Configure new subcloud region value
>   lineinfile:
>     path: "{{deploy_config}}"
>     regexp: 'OS_REGION_NAME'
>     line: "  OS_REGION_NAME: {{os_region_name_from_env.stdout}}"
>     backrefs: true
>     state: present
>   when: ("subcloud" in get_distributed_cloud_role.stdout) and os_region_name_exists.rc == 0
> ```

if we setting 'backrefs to true' to avoid grep, do we still rely on (os_region_name_exists.rc == 0) ?

sriram-gn:
> > Hi @yjian118 ,
> > I think by setting backrefs to true, we can avoid using `grep`.
> > Something like following snippet:
> > ```
> > - name: Configure new subcloud region value
> >   lineinfile:
> >     path: "{{deploy_config}}"
> >     regexp: 'OS_REGION_NAME'
> >     line: "  OS_REGION_NAME: {{os_region_name_from_env.stdout}}"
> >     backrefs: true
> >     state: present
> >   when: ("subcloud" in get_distributed_cloud_role.stdout) and os_region_name_exists.rc == 0
> > ```
> 
> if we setting 'backrefs to true' to avoid grep, do we still rely on (os_region_name_exists.rc == 0) ?

That's a good catch, sorry for the typo.
We will not be needing ` and os_region_name_exists.rc == 0` with `backrefs` set to `true`

yjian118:
@sriram-gn @sselvara1  Thanks for referring the backreference method, updated the backreference regex with sed, coveres the scenarios as:

In some cases, the deployment config doesn't contain the os region
name, the ansible lineinfile will append the os region name to the end
of the configuration file which result failed to apply to the cluster.

This commit:
1. Avoids unnecessary tasks to get the region name if not a subcloud
2. Uses sed instead of lineinfile:
   a. avoids adding new line if not match
   b. keeps spaces before OS_REGION_NAME to avoid format error if there
are four spaces before that.
   c. ensures all lines of the OS_REGION_NAME are updated(though
should not happen in most cases).

Test plan:
1. Deploy AIOSX stand alone, and update it with the playbook.
2. Deploy AIOSX subcloud, and update it by 'dcmanager subcloud deploy
config' with empty model.
3. Deploy AIOSX subcloud, and update it by 'dcmanager subcloud deploy
config' with full model.

sselvara1 (commented):
LGTM

sriram-gn:
@yjian118 just one clarification:
Should we restrict `OS_REGION_NAME` update only for the subclouds or should we keep it open for all controllers irrespective of DC role?
Wondering if AIO-SX standalone or other such non-DC systems could have `OS_REGION_NAME` other than `RegionOne`.
Apart from that, LGTM!

yjian118:
> @yjian118 just one clarification: Should we restrict `OS_REGION_NAME` update only for the subclouds or should we keep it open for all controllers irrespective of DC role? Wondering if AIO-SX standalone or other such non-DC systems could have `OS_REGION_NAME` other than `RegionOne`. Apart from that, LGTM!

No, the Non-DC and system controllers are using "  OS_REGION_NAME: RegionOne"

hugonicodemos (approved):

PR #321, opened 2024-01-26
Fix Incorrect PlatformNetwork Data Generated by Deployctl

sriram-gn:
This commit fixes incorrect PlatformNetwork data such as the platform
    network name and the platform network allocation type being generated by
    the deployctl tool.

    Test Plan:
    1. PASS: Build "deployctl" tool successfully.
    2. PASS: Generate deployment-config.yaml with correct data using
             deployctl.
    3. PASS: Verify that deployment-config.yaml generated with
             "--minimal-config" option does not contain any of OAM,
             management and admin platform networks.
    4. PASS: Verify that OAM, management and admin platform networks part of
             full deployment config when applied onto another system
             in bootstrap scope reflects accurate "InSync" status.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -1,5 +1,5 @@
 /* SPDX-License-Identifier: Apache-2.0 */
-/* Copyright(c) 2019-2023 Wind River Systems, Inc. */
+/* Copyright(c) 2019-2024 Wind River Systems, Inc. */
 
 package build
 
@@ -38,19 +38,21 @@ type Builder interface {
 	AddSystemFilters(filters []SystemFilter)
 	AddProfileFilters(filters []ProfileFilter)
 	AddHostFilters(filters []HostFilter)
+	AddPlatformNetworkFilters(filters []PlatformNetworkFilter)
 }
 
 // DeploymentBuilder is the concrete implementation of the builder interface
 // which is capable of building a full deployment model based on a running
 // system.
 type DeploymentBuilder struct {
-	client         *gophercloud.ServiceClient
-	namespace      string
-	name           string
-	progressWriter io.Writer
-	systemFilters  []SystemFilter
-	profileFilters []ProfileFilter
-	hostFilters    []HostFilter
+	client                 *gophercloud.ServiceClient
+	namespace              string
+	name                   string
+	progressWriter         io.Writer
+	systemFilters          []SystemFilter
+	profileFilters         []ProfileFilter
+	hostFilters            []HostFilter
+	platformNetworkFilters []PlatformNetworkFilter
 }
 
 var defaultSystemFilters = []SystemFilter{
@@ -169,6 +171,10 @@ func (db *DeploymentBuilder) AddProfileFilters(filters []ProfileFilter) {
 // on the deployment builder (if any).
 func (db *DeploymentBuilder) AddHostFilters(filters []HostFilter) {
 	db.hostFilters = append(db.hostFilters, filters...)
+}
+
+func (db *DeploymentBuilder) AddPlatformNetworkFilters(filters []PlatformNetworkFilter) {
+	db.platformNetworkFilters = append(db.platformNetworkFilters, filters...)
 }
 
 // Build is the main method which produces a deployment object based on a
@@ -561,12 +567,12 @@ func (db *DeploymentBuilder) buildPlatformNetworks(d *Deployment) error {
 		return err
 	}
 
-	nets := make([]*starlingxv1.PlatformNetwork, 0)
+	d.PlatformNetworks = make([]*starlingxv1.PlatformNetwork, 0)
 	always_generate_networks := []string{"storage", "mgmt", "oam", "admin"}
 	sort.Strings(always_generate_networks)
 	for _, p := range pools {
 		skip := false
-		network_type := networks.NetworkTypeOther
+		network := networks.Network{}
 		for _, n := range results {
 			if n.PoolUUID == p.ID {
 				// TODO(alegacy): for now we only support networks used for data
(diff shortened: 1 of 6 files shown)


sharbhar:
LGTM

sselvara1 (commented):
can we look at the "snippet" in deployment_config.yaml to see what changed after this commit?

sriram-gn:
> can we look at the "snippet" in deployment_config.yaml to see what changed after this commit?

Below is the diff @sselvara1:

```
diff deployment-config.yaml deployment-config-new.yaml
1,2c1,2
< # Generated: Wed Jan 31 11:21:01 UTC 2024
< # Tool version: v2.0.9 (master: e369ef5a66318156ece65c74c58ee54f046e9f0e)
---
> # Generated: Wed Jan 31 10:39:33 UTC 2024
> # Tool version: v2.0.9 (deployctl_platformnetwork_fix: 74902a0893519ff67422bf523a575b24d600fa76)
296c296
<   name: management
---
>   name: mgmt
325c325
<     type: dynamic
---
>     type: static
345c345
<     type: dynamic
---
>     type: static
```

The name is changed to "mgmt" and type to "static" in accordance with the system network.

sriram-gn:
@yjian118 I'm not sure if there's an easy way to determine whether the "controller0_address" and "controller1_address" were specified by the user or automatically assigned. The system API just provides the data either way.
Do you think having c0 and c1 ips anyway is a bad idea?

yjian118:
@sriram-gn Let's keep the topic of supporting the c0 and c1 addresses in another PR. Please filter out the 3 types of networks if minimal-config

sriram-gn:
@yjian118 checked-in changes wrt. "filter out the 3 types of networks if minimal-config".
Requesting you to kindly review the changes.

build/build.go
```diff
@@ -600,6 +606,11 @@ func (db *DeploymentBuilder) buildPlatformNetworks(d *Deployment) error {
 
 	d.PlatformNetworks = nets
 
+	err = db.filterPlatformNetworks(nil, d)
```
yjian118, line 609: Why are we offering a nil value?

yjian118 (changes requested):
Request unit test regarding the new filters

build/filter.go
```diff
@@ -724,6 +726,39 @@ func (in *DRBDLinkUtilizationFilter) Filter(system *v1.System, deployment *Deplo
...
+// This will filter out platformnetworks of type oam, mgmt and admin
+type CoreNetworkFilter struct {
+}
+
+func NewCoreNetworkFilter() *CoreNetworkFilter {
+	return &CoreNetworkFilter{}
+}
+
+func (in *CoreNetworkFilter) Filter(platform_network *v1.PlatformNetwork, deployment *Deployment) error {
+	if platform_network == nil {
```
yjian118, line 743: Isn't it always nil? What if it is not nil?

yjian118 (changes requested):
Need to update the test plan in the commit message, please paste a sample of output with or w/o minimal config

sriram-gn:
> Need to update the test plan in the commit message, please paste a sample of output with or w/o minimal config

The deployment config generated without `--minimal-config` contains:
```
---
apiVersion: starlingx.windriver.com/v1
kind: PlatformNetwork
metadata:
  labels:
    controller-tools.k8s.io: "1.0"
  name: mgmt
  namespace: deployment
spec:
  allocation:
    order: random
    ranges:
    - end: 192.168.204.50
      start: 192.168.204.2
    type: dynamic
  controller0Address: 192.168.204.3
  controller1Address: 192.168.204.4
  floatingAddress: 192.168.204.2
  prefix: 24
  subnet: 192.168.204.0
  type: mgmt
---
apiVersion: starlingx.windriver.com/v1
kind: PlatformNetwork
metadata:
  labels:
    controller-tools.k8s.io: "1.0"
  name: oam
  namespace: deployment
spec:
  allocation:
    order: random
    ranges:
    - end: 10.10.10.254
      start: 10.10.10.1
    type: static
  controller0Address: 10.10.10.3
  controller1Address: 10.10.10.4
  floatingAddress: 10.10.10.2
  gateway: 10.10.10.1
  prefix: 24
  subnet: 10.10.10.0
  type: oam

```

The whole section is absent in deployment config **with** `--minimal-config`.

Test plan has been updated in the commit message.

build/build.go
sriram-gn: We want to filter out core networks after having the complete list of deployment.PlatformNetworks.
Offering nil value signals CoreNetworkFilter.Filter can loop through once and for all for filtering out core networks.

yjian118:
Need to update the test in build_suite_test.go

build/filter.go
yjian118: If we want to introduce more filters, isn't it handled by newFilter.Filter()? Passing a new nil argument and set up a condition is quite weird. I understand it works because we have the "d.PlatformNetworks = nets" before, suggest:
1 Follow the existing filter's pattern: Filter(profile, deployment), assigning the networks to the deployment in/after the filter
or
2 remove the always nil argument and the condition

build/build_suite_test.go
```diff
@@ -1,5 +1,5 @@
 /* SPDX-License-Identifier: Apache-2.0 */
-/* Copyright(c) 2023 Wind River Systems, Inc. */
```
yjian118, line 2: 2023-2024

build/build.go
```diff
@@ -584,22 +589,28 @@ func (db *DeploymentBuilder) buildPlatformNetworks(d *Deployment) error {
...
+			// We are passing pointer to PlatformNetwork resource (net), hence, calling filterPlatformNetworks
+			// before or after append wouldn't make any difference and either way it would work.
+			d.PlatformNetworks = append(d.PlatformNetworks, net)
+			err = db.filterPlatformNetworks(net, d)
 			if err != nil {
 				return err
 			}
-			nets = append(nets, net)
 		}
 	}
```
yjian118, line 613: if skip, we lose the d.PlatformNetworks = nets, suggest to remain the previous logic

sriram-gn: fixed.

build/build_suite_test.go
sriram-gn: updated.

build/build.go
yjian118: If skip = true: d.PlatformNetworks = make([]*starlingxv1.PlatformNetwork, 0)

build/filter_test.go
```diff
@@ -1,5 +1,5 @@
 /* SPDX-License-Identifier: Apache-2.0 */
-/* Copyright(c) 2023 Wind River Systems, Inc. */
+/* Copyright(c) 2024 Wind River Systems, Inc. */
```
yjian118, line 2: 2023-2024

build/build.go
sriram-gn: I see a problem if we add it under `If skip = true` condition. 
What if we have `pools=['oam', 'mgmt', 'storage' ... , 'pxeboot']`:
The last one being `pxeboot` address pool, skip would be `true` and executing `d.PlatformNetworks = make([]*starlingxv1.PlatformNetwork, 0)` in that condition would erase all the networks added so far ('oam', 'mgmt', 'storage') which is not intended.

build/filter_test.go
sriram-gn: Fixed.

yjian118 (approved):

sselvara1 (approved):

PR #325, opened 2024-02-09
Avoid subcloud error collecting after the initial unlock

yjian118:
The error collecting is only designed in the initial deployment currently. The current logic relies on whether the "deploy_config" is offered is not correct as it can also be offered during reconfiguration.

This commit switches to use the initial_deploy flag, ensures the error reporting section will not be triggered after the initial deployment.

Test:
1. Passed - inital subcloud deployment successfully, verify the subcloud's error collecting during deployment.
2. Passed - reconfigure the subcloud, verify the subcloud's error collecting tasks are skipped.
3. Passed - lock a subcloud, reconfigure the subcloud's platform cores, verify the error collecting tasks are skipped.
4. Passed - Deploy a duplex subcloud and configure with DM
                  Swact the duplex subcloud controller-0
                  Reconfigure the duplex subcloud
                  Verify that error collection is skipped

### DIFF
--- a/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
+++ b/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-# Copyright(c) 2021-2023 Wind River Systems, Inc.
+# Copyright(c) 2021-2024 Wind River Systems, Inc.
 - name: Deployment Manager Playbook
   hosts: all
   gather_facts: false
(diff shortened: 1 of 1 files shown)


sselvara1:
I think title 'Avoid subcloud error collecting in reconfiguration'  gives the impression that subcloud error collecting is avoided altogether during reconfig.

can we change to - 'Avoid subcloud reconfig error due to wrong DM config' - to emphasis **when** we avoid error collecting. Thanks.

Else code looks good to me.

hugonicodemos (approved):

yjian118:
> I think title 'Avoid subcloud error collecting in reconfiguration' gives the impression that subcloud error collecting is avoided altogether during reconfig.
> 
> can we change to - 'Avoid subcloud reconfig error due to wrong DM config' - to emphasis **when** we avoid error collecting. Thanks.
> 
> Else code looks good to me.

It avoids the error collecting if the subcloud is after the initial deployment. I corrected the title, let me know if you have further questions/ concerns.

sriram-gn:
@yjian118 , can we add following test case:

- Deploy a duplex subcloud and configure with DM
- Swact the duplex subcloud controller-0
- Reconfigure the duplex subcloud
- Verify that error collection is skipped

yjian118:
> @yjian118 , can we add following test case:
> 
> * Deploy a duplex subcloud and configure with DM
> * Swact the duplex subcloud controller-0
> * Reconfigure the duplex subcloud
> * Verify that error collection is skipped
> 
> Wondering if this would work since we are checking a file and controller-1 might not be having it even though for the system itself it's not the initial unlock.

The /var/run/.nfv-vim.dor_complete is intended to be present in the active controller by vim. 
Just verified the test case you required.

sriram-gn:
Thanks @yjian118 . The commit LGTM.

sselvara1 (approved):

sriram-gn (approved):

PR #326, opened 2024-02-13
Change in DM helm-install to system app upload/apply

sharbhar:


yjian118 (changes requested):
Considering the current usage of this playbook, I'll not approve this change unless addressed the following concerns:
1 blocked initial deployment(should be done in bootstrap)
2 properly finished the upgrade/update condition
3 covers the usage of the day-2 operation

sharbhar:
abandoning this commit

sharbhar:
reopening

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -851,4 +903,4 @@
         - deploy_config is defined
         - "'subcloud' in get_distributed_cloud_role.stdout"
         - "'unlocked' not in current_administrativestate.stdout"
-        - "'unlocked' in administrative_state"
+        - "'unlocked' in administrative_state"
```
sriram-gn, line 906: Linefeed character is missing before the end of file (EOF).

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -309,16 +312,65 @@
...
-    - name: Install Deployment Manager
-      shell: KUBECONFIG=/etc/kubernetes/admin.conf /usr/sbin/helm upgrade --install deployment-manager {% if helm_chart_overrides is defined %}--values {{ helm_chart_overrides }}{% endif %} {{ manager_chart }}
-      when: helmv3_installed
-
-    # Install or Upgrade DM-Monitor
-    - name: "Install or Upgrade DM-monitor"
-      shell: "ansible-playbook {{ dm_monitor_playbook }}"
-      become: yes
-      ignore_errors: yes
-      when: dm_monitor_playbook
```
sriram-gn, line 321: If we are planning for backward compatibility of this playbook we cannot remove this.
Otherwise, we have to make sure all relevant tasks of helmv2 and helmv3 upgrade / installation is going to be handled by system app framework and such tasks be deleted from the playbook (tasks such as deleting webhook service, restarting DM pod etc.).

sriram-gn (changes requested):
Few more concerns:

1. What if helm chart is a user uploaded artifact? I don't see this being handled.
2. I see old path references for helm / overrides such as:
`/usr/local/share/applications/helm/wind-river-cloud-platform-deployment-manager-*.tgz`

The original playbook is designed to fail if it doesn't find these files in path and user uploaded artifacts is false.
 
Depending upon the plan for backward compatibility these things have to be handled.
Please add them to test cases as well.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -309,16 +312,65 @@
...
+    - name: Check deployment-manager App is present
+      shell: source /etc/platform/openrc; system application-list | grep deployment-manager | awk '{ print $10 }'
+      register: dm_app_status
+
+    # DM system app check, upload/apply helm-chart
+    - block:  
+        # Check dm-system app helm-chart
+       - name: Check DM system-app helm-chart present in /usr/local/share/applications/helm/
+         shell: ls {{ manager_app_chart }}
```
yjian118, line 323: Please use ansible.builtin.file to check the existence of a file

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -309,16 +312,65 @@
...
-      ignore_errors: yes
-      when: dm_monitor_playbook
+    - name: Check deployment-manager App is present
+      shell: source /etc/platform/openrc; system application-list | grep deployment-manager | awk '{ print $10 }'
+      register: dm_app_status
+
+    # DM system app check, upload/apply helm-chart
+    - block:  
+        # Check dm-system app helm-chart
+       - name: Check DM system-app helm-chart present in /usr/local/share/applications/helm/
```
yjian118, line 322: Can we use an ansible variable instead of hard-coding the file path? So we can let the StarlingX user overwrite it if required

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -309,16 +312,65 @@
...
+   
+       - block:
+          # dm-system app upload/apply
+         - name: Upload Deployment Manager System App Helm Chart
+           shell: source /etc/platform/openrc; system application-upload {{ manager_app_chart }}
+
+         - wait_for:
+           # Pause for an arbitrary amount of time to allow the deployment
+           # manager app to be uploaded.
+             timeout: 6
```
yjian118, line 340: Is 6 seconds enough to upload the tarball?

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
yjian118: agree

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -309,16 +312,71 @@
...
+         - name: Search Deployment Manager Apply Status
+           shell: source /etc/platform/openrc; system application-list | grep deployment-manager | awk '{ print $10 }'
+           register: app_apply_status
+
+         - name: Fail DM system app apply fails
+           fail:
+             msg: "The apply of DM system app has failed."
+           when: ("apply-failed" in app_apply_status.stdout)
+
+         when: (app_helm_chart.files[0].path | length != 0)
```
yjian118, line 377: This block & condition is not necessary as expected to fail before the start of the block if condition not met

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -309,16 +312,71 @@
 
       when: helmv3_installed and helmv3_dm_exists.rc == 0
 
-    - name: Install Deployment Manager
-      shell: KUBECONFIG=/etc/kubernetes/admin.conf /usr/sbin/helm upgrade --install deployment-manager {% if helm_chart_overrides is defined %}--values {{ helm_chart_overrides }}{% endif %} {{ manager_chart }}
-      when: helmv3_installed
```
yjian118, line 314: Need to re-visit all the tasks with "helmv3_installed" if the application no longer expected to be exist in helmv3 cli

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -309,16 +312,71 @@
...
+    - name: Check deployment-manager App is applied
+      shell: source /etc/platform/openrc; system application-list | grep deployment-manager | awk '{ print $10 }'
+      register: dm_app_status
 
-    # Install or Upgrade DM-Monitor
-    - name: "Install or Upgrade DM-monitor"
-      shell: "ansible-playbook {{ dm_monitor_playbook }}"
-      become: yes
-      ignore_errors: yes
-      when: dm_monitor_playbook
```
yjian118, line 321: Need to remove all dm_monitor_playbook related facts/tasks

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
yjian118: Also request to clean helmv2_installed as no longer supported

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -309,16 +312,71 @@
...
+    - name: Check if unlock is ready
+      stat:
+        path: /etc/platform/.unlock_ready
```
yjian118, line 321: What's the meaning of this flag? Can you add a comment to describe it?

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -309,16 +312,71 @@
...
+           register: app_apply_status
+
+         - name: Fail DM system app apply fails
+           fail:
+             msg: "The apply of DM system app has failed."
+           when: ("apply-failed" in app_apply_status.stdout)
+
+         when: (app_helm_chart.files[0].path | length != 0)
+
+      when: ((dm_app_status.stdout | length == 0) and (is_unlock_ready.stat.exists == True))
```
yjian118, line 379: What is the "is_unlock_ready" mean here? Why do we want to check it for this whole block? Can you add some comments above the block or here to describe that?

sharbhar:
Testing in progress.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -225,110 +183,53 @@
...
+         shell: source /etc/platform/openrc; system application-list | grep deployment-manager | awk '{ print $10 }'
+         register: app_apply_status
+         retries: 3
+         delay: 80
+         until: ("applied" in app_apply_status.stdout)
+
+       - name: Fail deployment-manager app apply fails
+         fail:
+           msg: "The apply of DM system app has failed."
+         when: ("apply-failed" in app_apply_status.stdout)
```
sriram-gn, line 230: It won't reach this task since the previous task won't succeed until `("applied" in app_apply_status.stdout)`.
We might have to ignore_errors in `Wait deployment-manager app to become applied` task with an update to fail task as shown below:

```
       - name: Fail if deployment-manager app apply fails
         fail:
           msg: "The apply of DM system app has failed. Expected 'applied' state but the app is in '{{app_apply_status.stdout}}' state."
         when: ("applied" not in app_apply_status.stdout)

```

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -225,110 +183,53 @@
...
+         when: ("uploaded" in app_upload_status.stdout)
+
+       - name: Wait deployment-manager app to become applied
+         shell: source /etc/platform/openrc; system application-list | grep deployment-manager | awk '{ print $10 }'
+         register: app_apply_status
+         retries: 3
+         delay: 80
+         until: ("applied" in app_apply_status.stdout)
+
+       - name: Fail deployment-manager app apply fails
```
sriram-gn, line 227: Please correct the name to: `name: Fail if deployment-manager app apply fails`

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -225,110 +183,53 @@
...
+         shell: source /etc/platform/openrc; system application-upload {{ manager_app_chart }}
+
+       - name: Wait deployment-manager app to become uploaded
+         shell: source /etc/platform/openrc; system application-list | grep deployment-manager | awk '{ print $10 }'
+         register: app_upload_status
+         retries: 3
+         delay: 6
+         until: ("uploaded" in app_upload_status.stdout)
+
+       - name: Fail deployment-manager app upload fails
```
sriram-gn, line 211: Please correct the name to `name: Fail if deployment-manager app upload fails`

sriram-gn (commented):
Please test some of the failure scenarios as well to make sure it works as expected.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -153,57 +149,19 @@
...
-      when: helmv3_installed
-
     - name: Check nfv-vim.dor_complete flag
       stat:
         path: /var/run/.nfv-vim.dor_complete
       register: dor_complete_file
 
+      # Check deployment-manager system app status
+    - name: Check deployment-manager app is status
+      shell: source /etc/platform/openrc; system application-list | grep deployment-manager | awk '{ print $10 }'
```
sriram-gn, line 159: Please use `--nowrap` with `system` commands otherwise it may be prone to unexpected failures.
`source /etc/platform/openrc; system application-list --nowrap | grep deployment-manager | awk '{ print $10 }'`

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -225,110 +183,53 @@
...
+         fail:
+           msg: "DM system app helm-chart does not exists in application directory"
+         when: (app_helm_chart.files[0].path | length == 0)
+   
+        # dm-system app upload/apply
+       - name: Upload deployment-manager app Helm Chart
+         shell: source /etc/platform/openrc; system application-upload {{ manager_app_chart }}
+
+       - name: Wait deployment-manager app to become uploaded
+         shell: source /etc/platform/openrc; system application-list | grep deployment-manager | awk '{ print $10 }'
```
sriram-gn, line 205: Please use `--nowrap` with `system` commands otherwise it may be prone to unexpected failures.
`source /etc/platform/openrc; system application-list --nowrap | grep deployment-manager | awk '{ print $10 }'`

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -225,110 +183,53 @@
...
+         fail:
+           msg: "The upload of deployment-manager system app has failed."
+         when: ("upload-failed" in app_upload_status.stdout)
+
+       - name: Apply deployment-manager app
+         shell: source /etc/platform/openrc; system application-apply deployment-manager   
+         when: ("uploaded" in app_upload_status.stdout)
+
+       - name: Wait deployment-manager app to become applied
+         shell: source /etc/platform/openrc; system application-list | grep deployment-manager | awk '{ print $10 }'
```
sriram-gn, line 221: Please use `--nowrap` with `system` commands otherwise it may be prone to unexpected failures.
`source /etc/platform/openrc; system application-list --nowrap | grep deployment-manager | awk '{ print $10 }'`

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -225,110 +183,59 @@
...
+         delay: 6
+         until: ("uploaded" in app_upload_status.stdout)
+         ignore_errors: yes
+
+       - name: Fail if deployment-manager app upload fails
+         fail:
+           msg: "The upload of deployment-manager system app has failed.
+                 Expected 'uploaded' state but the app is in 
+                 '{{ app_upload_status.stdout }}' state "
+         when: ("upload-failed" in app_upload_status.stdout)
```
sriram-gn, line 217: Fail if anything other than "uploaded".
`when: not ("uploaded" in app_upload_status.stdout)`

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -225,110 +183,59 @@
...
+         delay: 10
+         until: ("applied" in app_apply_status.stdout)
+         ignore_errors: yes
+
+       - name: Fail if deployment-manager app apply fails
+         fail:
+           msg: "The apply of DM system app has failed. 
+                 Expected 'applied' state but the app is in 
+                 '{{ app_apply_status.stdout }}' state."
+         when: ("apply-failed" in app_apply_status.stdout)
```
sriram-gn, line 236: Fail if anything other than "applied".
when: not ("applied" in app_apply_status.stdout)

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -5,17 +5,10 @@
   gather_facts: false
   become: false
 
-  vars:
-    helmv2_installed: false
-    helmv3_installed: false
-
   tasks:
     - set_fact:
         manager_chart: "{{ deployment_manager_chart | default('wind-river-cloud-platform-deployment-manager.tgz') }}"
```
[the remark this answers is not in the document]
sriram-gn: Unresolved this conversation since I did not see any code changes related to this, please respond to the query and then resolve it, thank you.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -153,57 +149,19 @@
...
-
     - name: Check nfv-vim.dor_complete flag
       stat:
         path: /var/run/.nfv-vim.dor_complete
       register: dor_complete_file
 
+      # Check deployment-manager system app status
+    - name: Check deployment-manager app is status
+      shell: source /etc/platform/openrc; system application-list --nowrap | grep deployment-manager | awk '{ print $10 }'
+      register: dm_app_status
```
yjian118, line 160: Do we need to verify the status? I assume "upload/upload-failed/applying/" should be treated differently

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -225,110 +183,59 @@
...
-    # Delete old Deployment Manager pod if it was reinstalled,
-    # the pod to delete may already be terminated by helm
-    - name: Restart Deployment Manager if reinstalled
-      command: >-
-        kubectl -n platform-deployment-manager delete pods {{ deployment_manager_pod_name.stdout }}
-      environment:
-        KUBECONFIG: "/etc/kubernetes/admin.conf"
-      when: deployment_manager_pod_name.stdout is defined and deployment_manager_pod_name.stdout
-      failed_when: false
+      when: not reconfig_flag
```
yjian118, line 238: May need to update the block for the following cases:

- application was uploaded - can we upload it again?
- application was upload-failed: I assume this block should be also fine in this case
- application was applying - should only waiting for applied
- application was removing:  I assume this block should be also fine in this case, suggest to double check
- application was remove-failed:  I assume this block should be also fine in this case, suggest to double check

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -225,110 +183,59 @@
...
-    # Install or Upgrade DM-Monitor
-    - name: "Install or Upgrade DM-monitor"
-      shell: "ansible-playbook {{ dm_monitor_playbook }}"
-      become: yes
-      ignore_errors: yes
-      when: dm_monitor_playbook
-
-    # Delete old Deployment Manager pod if it was reinstalled,
-    # the pod to delete may already be terminated by helm
-    - name: Restart Deployment Manager if reinstalled
```
yjian118, line 325: Expect to preserve this task, please check the pod's name and delete the pod in the case not a fresh install

yjian118:
Please consider all the application status need to handle during day-2 operation and update the procedures accordingly

sharbhar:
Testing completed.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -153,56 +149,27 @@
...
+       kubectl -n platform-deployment-manager get pods | grep platform-deployment-manager- | awk 'NR == 1 { print $3 }'
+      environment:
+          KUBECONFIG: "/etc/kubernetes/admin.conf"
+      register: deployment_manager_pod_status
+
     - name: Set reconfiguration related facts
       set_fact:
-        reconfig_flag: "{{ true if (helmv2_installed and helmv2_dm_exists.rc == 0)
-                        or (helmv3_installed and helmv3_dm_exists.rc == 0)
+        reconfig_flag: "{{ true if (dm_app_status.stdout == 'applied' and
```
yjian118, line 171: expect to handle the case the dm_app_status.stdout != 'applied' or deployment_manager_pod_status.stdout != 'Running', we cannot simply set the reconfig flag = False in these cases

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -153,56 +149,27 @@
...
       stat:
         path: /var/run/.nfv-vim.dor_complete
       register: dor_complete_file
 
+      # Check deployment-manager system app status
+    - name: Check deployment-manager app is status
+      shell: source /etc/platform/openrc; system application-list --nowrap | grep deployment-manager | awk '{ print $10 }'
+      register: dm_app_status
+
+    - name: Check deployment-manager pod is running
```
yjian118, line 162: Suggest using the same commands to wait for the pods to be ready below in concern with the pod in a transitional status

yjian118: Further are there any cases the application is 'applied' but pod is not 'Running'?  I assume this check should be included in the definition of 'applied', but I'm okay with this check.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -374,21 +299,23 @@
             deploy_config: "/home/{{ ansible_ssh_user }}/deployment-config.yaml"
           when: inventory_hostname != 'localhost'
 
-        - name: Configure required region value for subcloud
+        - name: Configure required region value if not systemcontroller
```
yjian118, line 302: I'm a bit concerned about why there's a change from the existing commit? Seems from a resolve of merge conflicts. We'll need to monitor the result after the merge

yjian118:
Please update the 1st message to include all the test cases passed with this commit

yjian118 (approved):
The change looks fine, is there any dependencies we need to care before merging the change?

sharbhar:
> The change looks fine, is there any dependencies we need to care before merging the change?
Checked, didn't find any dependencies to check.
we can merge.

sharbhar:
> Please update the 1st message to include all the test cases passed with this commit

Already present.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
sharbhar: Will monitor, once the result will be merged.

PR #328, opened 2024-02-15
Fix Host Lock Prevention During Admin Network Reconfiguration

sriram-gn:
This commit intends to fix the host lock execution prevention due to Host
CRD update.

After successful reconciliation of platform networks, the platform
network controller adds annotation on Host reconciler which triggers
network reconciliation of the host.
In this case, Host resource was modified by creating the annotation just
before host reconciler was set to send lock_required strategy update in
order to create new admin VLAN interface.
As a consequence of this, lock_required strategy update failed due to
request target being obselete verion of Host CRD.

To address this issue, annotation as a way to signal host reconciler for
network reconciliation post platform network reconciliation is removed.
Instead we are setting host reconciliation status to 'false' and letting
host reconciler handle rest of it.

Test Cases:
1. PASS - Verify network reconfig is successful on AIO-SX systems by
          updating platform resource configurations with deploymentScope
          of "principal" for oam/mgmt/admin networks.
2. PASS - Verify network reconfig is successful on AIO-SX systems by
          applying new platform resource configurations with
          deploymentScope of "principal" for oam/mgmt/admin networks.
3. PASS - Verify that platform network reconciliation and host
          reconciliation are successful when new admin interface is
          created.
4. PASS - Verify request with emtpy opts is not sent to sysinv from
          gophercloud when ptpRole attribute is absent in the Host /
          HostProfile.

### DIFF
--- a/controllers/host/host_controller.go
+++ b/controllers/host/host_controller.go
@@ -694,7 +694,8 @@ const MinimumEnabledControllerNodesForNonController = 2
 func (r *HostReconciler) ReconcileFinalState(client *gophercloud.ServiceClient, instance *starlingxv1.Host, profile *starlingxv1.HostProfileSpec, host *v1info.HostInfo) error {
 	state := profile.AdministrativeState
 	if state == nil || *state == host.AdministrativeState ||
-		instance.Status.DeploymentScope == cloudManager.ScopePrincipal {
+		instance.Status.DeploymentScope == cloudManager.ScopePrincipal ||
+		instance.Status.StrategyRequired != cloudManager.StrategyNotRequired {
 		// No action required.
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


controllers/host/host_controller.go
```diff
@@ -1064,10 +1065,15 @@ func (r *HostReconciler) CompareDisabledAttributes(in *starlingxv1.HostProfileSp
 func (r *HostReconciler) ReconcileHostByState(client *gophercloud.ServiceClient, instance *starlingxv1.Host, current *starlingxv1.HostProfileSpec, profile *starlingxv1.HostProfileSpec, host *v1info.HostInfo) error {
 
 	principal := false
+	requires_vim_strategy := false
```
yjian118, line 1068: Can you add a comment above regarding this variable? The name is confusing

api/v1/hostprofile_types.go
```diff
@@ -435,7 +435,7 @@ type CommonInterfaceInfo struct {
 
 	// PTPRole defines the ptp role as master, slave, or none
 	// +kubebuilder:validation:Enum=master;slave;none
-	PTPRole *string `json:"ptpRole,omitempty"`
```
yjian118, line 438: I would not prefer to request the user to update their existing deployment-config.yaml for a none value caused by adding mandatory CRD, though it is the simplest approach to the developers.
Can we check&fix it during composing the default& final profiles?

yjian118 (changes requested):
I would hold this review due to the concern of updating the deployment-config due to mandatory PTPRole. This commit is risky to merge as requesting lab update. @sriram-gn can you fix the PTPRole with another approach? The rest change looks good

hugonicodemos (changes requested):
Can we add a test at least for the new method?

sriram-gn:
> Can we add a test at least for the new method?

@hugonicodemos the new method involves gophercloud and to write UT for any method that involves gophercloud requires some exploration to do. Requesting UT for this to be tracked in a separate PR.

controllers/host/host_controller.go
sriram-gn: Done.

yjian118: Suggest to rename this variable to "strategy_required"

api/v1/hostprofile_types.go
sriram-gn: No updates requried here.
`MergeProfiles(defaults, profile)` already handles it.

controllers/host/host_controller.go
sriram-gn: > Suggest to rename this variable to "strategy_required"

Done.

hugonicodemos (approved):

sselvara1 (approved):
LGTM with all testing covered.

yjian118 (approved):

PR #333, opened 2024-03-06
Extend timeout for duplex and multi-nodes subclouds

yjian118:
Extend timeout for duplex and multi-nodes subclouds
This commit extends the timeout when deploying a duplex and
multi-nodes subclouds:

1. Check the number of hosts from the cluster
2. Check the availabilityStatus of the hosts, fail the further
operation if any host remain offline due to problem during early
installation.
3. Extend timeout of waiting for the resources to be reconciled based
on the hosts numbers.

Test plan:
1. Deploy a DC with OSDs in a standard, AIOSX, AIODX subclouds.
2. Reconfig the subclouds with empty model.
3. Reconfig the MTU of the subclouds.

### DIFF
--- a/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
+++ b/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
@@ -717,30 +717,81 @@
 
             when: waiting_after_reboot.failed
 
-          # After reboot, wait some time for resources to be reconciled.
-          # If we have all of them reconciled, playbook won't fail.
-          # This task is getting the last column as RECONCILED status value
-          # TODO(ecandotti): Modify this block to not retrieve the last column, but instead recognize the column number for RECONCILED in each resource.
-          # TODO(ecandotti): This will eliminate the need to always have the RECONCILED column as the last column when a new state is added.
-          - name: Retrieve kubectl resources reconciled status
-            shell: >-
-              (kubectl -n deployment get datanetworks,platformnetworks,systems,ptpinstances,ptpinterfaces;
-               kubectl -n deployment get hosts "{{ get_host_name.stdout}}") |
-               awk '$NF ~ /^false/ {print}'
+          - name: Set retries to check resource reconciled status for simplex
+            set_fact:
+              host_reconcile_retries: 80
+
+          - name: Check number of hosts from deployment namespace
+            shell: kubectl get hosts -n deployment |  grep -v NAME  | wc -l
+            environment:
+              KUBECONFIG: "/etc/kubernetes/admin.conf"
+            register: hosts_count
+            until: hosts_count.stdout != 0
+            retries: 20
+            delay: 30
+            failed_when: false
+
+          - name: Fail to get hosts from cluster
+            fail:
+              msg: "Failed to get host resource from cluster."
+            when: hosts_count.stdout == 0
+
+          - name: Wait for subcloud hosts available if not simplex
+            block:
+            - name: Set system type for monitor the deloyment
+              set_fact:
+                host_offline_retries: >
+                  {{ 60 if hosts_count.stdout | int == 2 else 120 }}
+                host_reconcile_retries: >
+                  {{ 120 if  hosts_count.stdout | int == 2 else 240 }}
+
+            - name: Get offline hosts if not simplex
+              shell: >
+                kubectl get hosts -o custom-columns='NAME:.metadata.name,
+                OPERATIONAL:.status.availabilityStatus' -n deployment |
+                awk '$NF ~ /offline/ {print}'
+              environment:
+                KUBECONFIG: "/etc/kubernetes/admin.conf"
+              register: get_offline_hosts
+              until: >
+                (get_offline_hosts.stdout == "" and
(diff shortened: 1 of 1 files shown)


sriram-gn:
Since we are checking the last column, I think it's safer to list values for only `.status.reconciled`.
I mean something like this:
```
 (kubectl -n deployment get datanetworks,platformnetworks,ptpinstances,ptpinterfaces -o custom-columns=RECONCILED:.status.reconciled;
               kubectl -n deployment get hosts "{{ get_host_name.stdout}}" -o custom-columns=RECONCILED:.status.reconciled) |
               awk '$NF ~ /^false/ {print}'

```

Current implementation wouldn't work for PlatformNetwork since it's having deploymentScope as the last column instead:
https://github.com/Wind-River/cloud-platform-deployment-manager/blob/master/config/crd/bases/starlingx.windriver.com_platformnetworks.yaml#L38

Bootstrap configuration for oam/mgmt and admin is ignored anyway, so it wouldn't affect for the most part as of now.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -722,9 +722,11 @@
           # This task is getting the last column as RECONCILED status value
           # TODO(ecandotti): Modify this block to not retrieve the last column, but instead recognize the column number for RECONCILED in each resource.
           # TODO(ecandotti): This will eliminate the need to always have the RECONCILED column as the last column when a new state is added.
-          - name: Retrieve kubectl resources reconciled status
+          # TODO(yuxing): Remove the monitor of the system resource for now, need to add it back and calculate the timeout
+          # based on the number of nodes and their inv status.
+          - name: Get not reconciled resource
             shell: >-
-              (kubectl -n deployment get datanetworks,platformnetworks,systems,ptpinstances,ptpinterfaces;
+              (kubectl -n deployment get datanetworks,platformnetworks,ptpinstances,ptpinterfaces;
```
sriram-gn, line 729: Since we are checking the last column, I think it's safer to list values for only `.status.reconciled`.
I mean something like this:
```
 (kubectl -n deployment get datanetworks,platformnetworks,ptpinstances,ptpinterfaces -o custom-columns=RECONCILED:.status.reconciled;
               kubectl -n deployment get hosts "{{ get_host_name.stdout}}" -o custom-columns=RECONCILED:.status.reconciled) |
               awk '$NF ~ /^false/ {print}'

```

Current implementation wouldn't work for PlatformNetwork since it's having deploymentScope as the last column instead:
https://github.com/Wind-River/cloud-platform-deployment-manager/blob/master/config/crd/bases/starlingx.windriver.com_platformnetworks.yaml#L38

Bootstrap configuration for oam/mgmt and admin is ignored anyway, so it wouldn't affect for the most part as of now.

yjian118: Thanks, updated as suggested, printed the resource name for the next tasks for debugging purpose

hugonicodemos (approved):

sriram-gn (approved):
LGTM

yjian118:
@hugonicodemos @sriram-gn since I re-wrote this commit, the change will be far different from the previous one you reviewed, can you help to review this change again? Thanks

hugonicodemos:
LGTM, @sriram-gn @sselvara1 please review it.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -717,30 +717,81 @@
...
+                host_offline_retries: >
+                  {{ 60 if hosts_count.stdout | int == 2 else 120 }}
+                host_reconcile_retries: >
+                  {{ 120 if  hosts_count.stdout | int == 2 else 240 }}
+
+            - name: Get offline hosts if not simplex
+              shell: >
+                kubectl get hosts -o custom-columns='NAME:.metadata.name,
+                OPERATIONAL:.status.availabilityStatus' -n deployment |
+                awk '$NF ~ /offline/ {print}'
```
sriram-gn, line 752: Wondering if we have to capture hosts that are in `online` or `power-off` state as well here and fail the playbook if hosts were found to be in any of these states `offline`, `online` and `power-off` after exhausting `host_offline_retries`.

I mean something like:
`awk '$NF ~ /offline/ || $NF ~ /online/ || $NF ~ /power-off/ {print}'`

Especially if the host is `locked-disabled-online` host we might jump to `Get not reconciled resource` a bit too early.

yjian118: The reason why I don't catch the "online" is because it takes a lot of time to wait for the unlock of a host, and then unlock another host...The 1 hour time out will not be enough to wait for all hosts to be unlocked, this task should be handled by the later reconciled check because the unlock was triggered during reconciliation.

This check only checks any host stay "offline" caused by the installation failure, and fails the playbook in advance.

Regarding the "power-off", the host is not expected to be power-off during the initial deployment, same as the controller-0 should not be marked as "locked"

PR #334, opened 2024-03-11
Patch back to MR2+

yjian118:
This PR added a list of commits that want to be patched back to MR2+, not straight cherry-picks, please help me to verify if any commit missed

PR #335, opened 2024-03-21
Fill empty IF uuid from config

yjian118:
This commit updates the interface UUID in the profile if it is empty by default.

I observed this case in a lab after upgrading from 21.12 to 22.12. As the interface UUID is recently introduced in the CRD, it is possible it is empty in host instance default profile and will not be updated after killing the pod(as the default is already generated).

As the deployctl avoids to generated the UUID, I turned to hack the deployctl, generate the UUID and apply to the cluster to fix the profile and make the host in-sync.

This change search the empty uuid with the interface name and fill it again from the current config, ensures the uuids not empty in default profiles. I made this change generic, but currently only observed this problem after platform upgrade.

Test-plan:
1 Deploy 21.12 AIOSX, upgrade the platform to 22.12 latest patch 2 Check the interface UUID in host default is updated with the UUIDs 3 Generate the deployment config and update the MTC as day-2 operation, verify the host can be insync and reconciled after update.


yjian118:
Test pending

controllers/host/profile_utils.go
```diff
@@ -572,3 +575,64 @@ func (r *HostReconciler) ValidateProfile(host *starlingxv1.Host, profile *starli
...
+	if profile == nil || current == nil {
+		return
+	}
+
+	if_current := current.Interfaces.Ethernet
+	if_profile := profile.Interfaces.Ethernet
+	for idx_profile := range if_profile {
+		info_profile := if_profile[idx_profile].CommonInterfaceInfo
+		profile_uuid := info_profile.UUID
+		if profile_uuid == "" || profile == nil {
```
sriram-gn, line 624: I think `|| profile == nil` should be removed from here.
If `profile == nil` it would return at line 616 and doesn't come this far though.

controllers/host/profile_utils.go
```diff
@@ -572,3 +575,64 @@ func (r *HostReconciler) ValidateProfile(host *starlingxv1.Host, profile *starli
...
+				if_profile[idx_profile].Port.Name = port_current
+				break
+			}
+		}
+	}
+}
+
+// FillEmptyUuidbyName fills the empty uuid by its name. Normally, the uuids can
+// be searched from the system and merged to the default profile.
+func FillEmptyUuidbyName(
```
sriram-gn, line 612: Does `SyncUuidByIFName` sound better?
Since we have `SyncIFNameByUuid`, right away we will know this is reverse of it.

yjian118: It's not sync, just fill it if missed

controllers/host/profile_utils.go
yjian118: Acked

yjian118: I think that's a mistake that I need to check if the "profile_uuid" is nil

sselvara1 (commented):
LGTM. Test cases need to be covered for PASS criteria.

yjian118:
Test done

yjian118:
Add support of VF, VLAN, Bond interface.
Verified fix on an AIOSX after platform upgrade with empty uuid of vlan, ethernet interfaces

yjian118:
> LGTM. Test cases need to be covered for PASS criteria.

Added

yjian118:
> Should we cherry-pick this commit here? #334

I want to, but depends on the workflow, I'm requesting that now

PR #343, opened 2024-04-03
Upversion v2.0.10-11

hugonicodemos:


yjian118 (commented):
Why do we need to update the chart? Any update of the CRDs?

PR #348, opened 2024-04-15
Update CertificateInfo in CRD to reject cert types other than ssl_ca

vineela-pachchipulusu:
Updating CertificateInfo in CRD to reject cert types other than ssl_ca

DM should no longer check any platform cert other than ssl_ca.
Remove the other types of certs from CRD

Test plan:

1. PASS:Day2 operation on AIO-SX VDM VM
image
2. PASS:DM upgrade with the deploy config with deprecated certificate types
image


sriram-gn (changes requested):
>delete other crt types except ssl_ca from CRD

Commit message should be thorough with appropriate title, sufficient description and test cases executed by developer to justify that changes made to the code actually works as intended.

sriram-gn (commented):
I think it might be risky to remove rest of the certificates from webhook validation without handling it in DM first.
The `UpdateConfigStatus` might error out if during the upgrade the system already has certificates such as ssl, openstack, openstack_ca etc.

vineela-pachchipulusu:
> I think it might be risky to remove rest of the certificates from webhook validation without handling it in DM first. The `UpdateConfigStatus` might error out if during the upgrade the system already has certificates such as ssl, openstack, openstack_ca etc.

This is handled by filtering deprecated cert types before triggering `UpdateConfigStatus` func

controllers/system/system_controller.go
```diff
@@ -1581,6 +1581,30 @@ func (r *SystemReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+	// Filter out certificates with type other than "ssl_ca"
+	var originalLength int
+	if instance.Spec.Certificates != nil {
+		originalLength = len(*instance.Spec.Certificates)
+		filteredCerts := make([]starlingxv1.CertificateInfo, 0, originalLength)
+		for _, cert := range *instance.Spec.Certificates {
+			if cert.Type == "ssl_ca" {
+				filteredCerts = append(filteredCerts, cert)
+			} else {
+				logSystem.Info("Removed certificates with all other deprecated types expect ssl_ca", "system", instance.Name)
```
sriram-gn, line 1593: Typo to be corrected, expect -> except.

sriram-gn (changes requested):

controllers/system/system_controller.go
vineela-pachchipulusu: Done.

sriram-gn (approved):

controllers/system/system_controller.go
```diff
@@ -1581,6 +1581,30 @@ func (r *SystemReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 
 	// Update scope from configuration
 	logSystem.V(2).Info("before UpdateConfigStatus", "instance", instance)
+	// Filter out certificates with type other than "ssl_ca"
+	var originalLength int
+	if instance.Spec.Certificates != nil {
+		originalLength = len(*instance.Spec.Certificates)
+		filteredCerts := make([]starlingxv1.CertificateInfo, 0, originalLength)
+		for _, cert := range *instance.Spec.Certificates {
+			if cert.Type == "ssl_ca" {
```
yjian118, line 1590: 1 Can we have the string "ssl_ca" replaced with PlatformCACertificate?
2 Would prefer to have this new section as a new method "clean_deprecated_certificates"
3 Expect a new test against the new method

controllers/system/system_controller_test.go
```diff
@@ -50,4 +50,48 @@ var _ = Describe("System controller", func() {
...
+	for _, cert := range *certs {
+		if cert.Type == starlingxv1.PlatformCACertificate {
+			filteredCerts = append(filteredCerts, cert)
+		} else {
+
+		}
+	}
+	return filteredCerts
+}
+*/
```
sriram-gn, line 97: Remove unwanted comments / code snippets.

sriram-gn (changes requested):

controllers/system/system_controller_test.go
vineela-pachchipulusu: Done.

controllers/system/system_controller.go
vineela-pachchipulusu: Implemented the `clean_deprecated_certificates` func and added  test case to the method

sriram-gn (approved):

yjian118 (approved):

PR #350, opened 2024-04-22
Update CRD(s) To Support Dual-stack Networking

sriram-gn:
In this commit we are mainly covering following to support dual-stack
network reconciliation workflows:

1. Update of PlatformNetwork CRD
   Here, we are removing all the address pool related parameters from
   the PlatformNetwork CRD and adding two new parameters:

   a. associatedAddressPools - List of AddressPool object names to
      associate the platform network.

   b. ipAllocationType - This can be enumeration of ("static",
      "dynamic"). It used to be part of AddressPool AllocationInfo.

   and renaming "type" to "networkType" so that it's clear what "type"
   we are talking about.

   Apart from this, we are also removing attributes such as
   `StrategyRequired` and `PoolUUID` from PlatformNetworkStatus since
   it's no longer valid to have them.

2. Creation of New AddressPool CRD
   A new CRD has been created to define address pool objects rather than
   having them under one single banner of PlatformNetwork CRD.
   This would not only align DM PlatformNetwork and AddressPool CRDs
   with StarlingX models but also support dual-stack configurations and
   could potentially support shared network design.

   Note: This CRD doesn't support `deploymentScope` attribute, it would
         be derived from PlatformNetwork objects to avoid ending up with
         partially reconfigured setups.

3. Creation of New AddressPool Reconciler
   The reconciler for AddressPool has been created such that it can
   accept the AddressPool objects and set controller-0's (hardcorded)
   InSync/Reconciled status to false so that host controller can take it
   forward.

   Note: The implementation will be updated later to notify active
         host controller instead of hardcoded controller-0 and use
         Generation/ObservedGeneration to determine if at all host
         controller has to be notified.

4. Update of PlatformNetwork Reconciler
   The current PlatformNetwork reconciler has been cut down for the most
   part to just include the behaviour of notifying the host controller
   if there are any changes to PlatformNetwork resources.
   Just as in AddressPool reconciler the PlatformNetwork reconciler will
   also be revisited later.

5. Creation of New PlatformNetwork Sub-Reconciler
   A sub-reconciler is created under host reconciler to handle all of
   platform network reconciliation workflows.
   In current commit the sub-reconciler just sets Reconciled/InSync
   statuses of all AddressPool/PlatformNetwork objects to true.
   It will be updated to handle all of Day-1 and Day-2 workflows of
   platform network reconciliation.

6. AddressPool Webhook Validations
   Validations have been updated for AddressPool object and unit test
   cases have been written to cover all the scenarios (~96% coverage).

7. Update of go.mod and go.sum to Enable Usage of Dual-Stack API(s)
   Gophercloud is already having changes related to new API(s) which
   needs to be leveraged to support dual-stack workflows.
   Hence, go.mod and go.sum have been updated to enable usage of new
   API(s).

8. Update of Deployment Manager Playbook
   DM playbook has been updated to check reconciliation status of
   address pool objects as well.
   This would fail the play if reconciliation of any of the address
   pools are unsuccessful.

    Test Cases:
    1. PASS: Verify build is executed successfully.
    2. PASS: Verify CRDs generated as expected.
    3. PASS: Verify PlatformNetwork & AddressPool objects with new spec can be
             applied successfully.
    4. PASS: Verify InSync/Reconciled status of  PlatformNetwork & AddressPool
             objects are true in AIO-SX.
    5. PASS: Verify all the unit test cases are executed successfully.
    6. PASS: Deploy system without this change and upgrade DM with this
             change.
    7. PASS: Build new deployctl and generate deployment configurations with
         / without "--minimal-config" option successfully.


api/v1/platformnetwork_webhook.go
```diff
@@ -42,56 +33,6 @@ func (r *PlatformNetwork) Default() {
...
-	}
-	return false
-}
-
-// Validates an incoming resource update/create request.  The intent of this validation is to perform only the
-// minimum amount of validation which should normally be done by the CRD validation schema, but until kubebuilder
-// supports the necessary validation annotations we need to do this in a webhook.  All other validation is left
-// to the system API and any errors generated by that API will be reported in the resource status and events.
-// Validates that all address specifications within the network are of the same address family.
-func (r *PlatformNetwork) validatePlatformNetwork() error {
```
yjian118, line 69: The removal of validatePlatformNetwork is unpleasant. However, the original validation only validates the address pool values. I'm fine with it. May need to consider if other validations, can you add (TODO) in the following ValidateCreate ValidateUpdate methods

sriram-gn: It has been added [here](https://github.com/Wind-River/cloud-platform-deployment-manager/pull/350/files#diff-5ccf80dc481f9473df3b4d9364346c7b872893b7002e4c10ff0bc4dc7e508eaeR41
) and [here](https://github.com/Wind-River/cloud-platform-deployment-manager/pull/350/files#diff-5ccf80dc481f9473df3b4d9364346c7b872893b7002e4c10ff0bc4dc7e508eaeR50).

controllers/platformnetwork_controller_fixtures.go
```diff
@@ -319,165 +315,165 @@ const DummyOAMUpdateResponse = `
...
-	case http.MethodPatch:
-		w.Header().Add("Content-Type", "application/json")
-		fmt.Fprint(w, DummyAddressPoolUpdateResponse)
-	case http.MethodDelete:
-		w.WriteHeader(http.StatusNoContent)
-	default:
-		http.Error(w, `{"error": "Method not allowed"}`, http.StatusMethodNotAllowed)
-	}
-}
+// func HandleAddressPoolRequests(w http.ResponseWriter, r *http.Request) {
```
yjian118, line 318: Any consideration to comment these methods than delete them?

config/crd/bases/starlingx.windriver.com_platformnetworks.yaml
```diff
@@ -16,16 +16,12 @@ spec:
   versions:
   - additionalPrinterColumns:
     - description: The platform network type.
-      jsonPath: .spec.type
-      name: type
+      jsonPath: .spec.networkType
```
yjian118, line 19: Any consideration to those specs printed out?

config/crd/bases/starlingx.windriver.com_platformnetworks.yaml
```diff
@@ -150,10 +94,9 @@ spec:
                 - other
                 type: string
             required:
-            - allocation
-            - prefix
-            - subnet
-            - type
+            - associatedAddressPools
+            - ipAllocationType
```
yjian118, line 98: to align with the concept in the sysinv, can we make it to "dynamic", and use boolean value?

api/v1/constructors.go
```diff
@@ -1437,30 +1437,10 @@ func NewPlatformNetworkSpec(pool addresspools.AddressPool, network networks.Netw
...
-		FloatingAddress:    pool.FloatingAddress,
-		Controller0Address: pool.Controller0Address,
-		Controller1Address: pool.Controller1Address,
-		Prefix:             pool.Prefix,
-		Gateway:            pool.Gateway,
-		Allocation: AllocationInfo{
-			Type:  allocation_type,
-			Order: &pool.Order,
-		},
+		NetworkType:      network.Type,
```
yjian118, line 1440: Do we need to construct the new address pool spec?

yjian118 (commented):
Expect test cases:
1 The deployctl.
2 deploy an system w/o this change, upgrade the dm with this change

api/v1/constructors.go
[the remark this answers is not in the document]
sriram-gn: Just that it will be clearer what `type` we are talking about.
I think we could still go ahead with `type` instead of `networkType` now that PlatformNetwork resource doesn't mix up AddressPool, what do you suggest @yjian118 ?

controllers/platformnetwork_controller_fixtures.go
sriram-gn: For platformnetwork_controller most of it will not be applicable since reconciliation will be moved to platformnetwork sub-reconciler.
We can refer to these fixtures and methods step by step while writing UT for sub-reconciler which will be in a different file, that way we will have only what's required for sub-reconciler.

api/v1/constructors.go
sriram-gn: Added constructor for AddressPool spec.
Had kept it for deployctl commit but we now have it anyway, PlatformNetwork constructor will be updated during deployctl commit to support exporting associatedAddressPools though.

controllers/addresspool_controller.go
```diff
@@ -0,0 +1,180 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+/* Copyright(c) 2019-2024 Wind River Systems, Inc. */
```
yjian118, line 2: 2024

api/v1/platformnetwork_types.go
```diff
@@ -7,62 +7,19 @@ import (
...
-	// +kubebuilder:validation:Maximum=128
-	Prefix int `json:"prefix"`
-
-	// Gateway defines the nexthop gateway IP address if applicable
-	// +optional
-	Gateway *string `json:"gateway,omitempty"`
-
-	// Allocation defines the allocation scheme details for the network
-	Allocation AllocationInfo `json:"allocation"`
+	// AssociatedNetworks defines the set of network names to associate
```
hugonicodemos, line 20: AssociatedNetworks or AssociatedAdressPools?

api/v1/platformnetwork_webhook.go
hugonicodemos: Why not validate the associated networks?

hugonicodemos (commented):
Couple comments in line.

controllers/addresspool_controller.go
sriram-gn: Fixed.

api/v1/platformnetwork_types.go
sriram-gn: Good catch!
It is "AssociatedAddressPools", updated the doc as well.

api/v1/platformnetwork_webhook.go
sriram-gn: @hugonicodemos currently we don't have any special validations in mind for platformnetworks other than kubebuilder validations that are in place.
As @yjian118 has pointed out, I've added a TODO on myself to identify such validations to be done if at all any.

PR #351, opened 2024-04-23
Update system reconciler to check the cert type and report error on unexpected certs

vineela-pachchipulusu:
Update system reconciler to check the cert type and throw log message on unexpected certs

DM should no longer check any platform cert other than ssl_ca. In system reconciler, we need to check the input and throw log message message on any unexpected cert type.

Test plan:

Day2 operation on AIO-SX VDM VM
DM upgrade

### DIFF
--- a/api/v1/system_webhook.go
+++ b/api/v1/system_webhook.go
@@ -146,6 +146,7 @@ func validateCertificates(obj *System) error {
 			// - Openstack_CA/OpenLDAP/Docker/SSL(HTTPS)
 			if c.Type == OpenstackCACertificate || c.Type == OpenLDAPCertificate ||
 				c.Type == DockerCertificate || c.Type == PlatformCertificate {
+				systemlog.Info("unexpected certificate", "type", c.Type)
 				continue
 			}
 


api/v1/system_webhook.go
```diff
@@ -146,7 +146,8 @@ func validateCertificates(obj *System) error {
 			// - Openstack_CA/OpenLDAP/Docker/SSL(HTTPS)
 			if c.Type == OpenstackCACertificate || c.Type == OpenLDAPCertificate ||
 				c.Type == DockerCertificate || c.Type == PlatformCertificate {
-				continue
+				msg := fmt.Sprintf("Unexpected certificate type: %s", c.Type)
+				return errors.New(msg)
```
sriram-gn, line 150: I would prefer to use kubebuilder enum validation for this instead but we cannot make this change until we have an answer for behavior of the DM when certs such as openstack_ca / docker / openldap / platform are applied in version N-1 and upgraded to N.

vineela-pachchipulusu:
This is handled in the https://github.com/Wind-River/cloud-platform-deployment-manager/blob/master/controllers/system/system_controller.go#L1227

PR #352, opened 2024-04-23
Reboot issue when insync is false

MahalakshmiManthiram:
- Added annotation restore in progress
- insync, reconcile status and deployment scope are captured in annotation
- Updated status in platform network contorller

### DIFF
--- a/controllers/datanetwork_controller.go
+++ b/controllers/datanetwork_controller.go
@@ -524,6 +524,22 @@ func (r *DataNetworkReconciler) Reconcile(ctx context.Context, request ctrl.Requ
 		return reconcile.Result{}, err
 	}
 
+	// Restore the data network status
+	if r.checkRestoreInProgress(instance) {
+		r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Restoring '%s' Data network resource status without doing actual reconciliation", instance.Name)
+		if err := r.RestoreDataNetworkStatus(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		if err := r.ClearRestoreInProgress(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation && instance.Status.Reconciled {
+		return ctrl.Result{}, nil
+	}
+
 	// Update scope from configuration
 	logDataNetwork.V(2).Info("before UpdateConfigStatus", "instance", instance)
 	err = r.UpdateConfigStatus(instance)
@@ -591,4 +607,62 @@ func (r *DataNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&starlingxv1.DataNetwork{}).
 		Complete(r)
+}
+
+// Verify whether we have annotation restore-in-progress
+func (r *DataNetworkReconciler) checkRestoreInProgress(instance *starlingxv1.DataNetwork) bool {
+	restoreInProgress, ok := instance.Annotations[cloudManager.RestoreInProgress]
+	if ok && restoreInProgress != "" {
+		return true
+	}
+	return false
+}
+
+// Update status
+func (r *DataNetworkReconciler) RestoreDataNetworkStatus(instance *starlingxv1.DataNetwork) error {
+	annotation := instance.GetObjectMeta().GetAnnotations()
+	config, ok := annotation[cloudManager.RestoreInProgress]
+	if ok {
+		restoreStatus := &cloudManager.RestoreStatus{}
+		err := json.Unmarshal([]byte(config), &restoreStatus)
+		if err == nil {
+			if restoreStatus.InSync != nil {
+				instance.Status.InSync = *restoreStatus.InSync
+			}
+			instance.Status.Reconciled = true
+			instance.Status.ObservedGeneration = instance.ObjectMeta.Generation
+			instance.Status.DeploymentScope = "bootstrap"
+			instance.Status.StrategyRequired = "not_required"
+			err = r.Client.Status().Update(context.TODO(), instance)
+			if err != nil {
+				log_err_msg := fmt.Sprintf(
+					"Failed to update data network status while restoring '%s' resource. Error: %s",
+					instance.Name,
+					err)
+				return common.NewResourceStatusDependency(log_err_msg)
+			} else {
+				StatusUpdate := fmt.Sprintf("Status updated for DataNetwork resource '%s' during restore with following values: Reconciled=%t InSync=%t DeploymentScope=%s",
+					instance.Name, instance.Status.Reconciled, instance.Status.InSync, instance.Status.DeploymentScope)
+				r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, StatusUpdate)
+
+			}
+		} else {
+			r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Failed to unmarshal '%s'", err)
+		}
+	}
+	return nil
+}
+
+// Clear annotation RestoreInProgress
+func (r *DataNetworkReconciler) ClearRestoreInProgress(instance *starlingxv1.DataNetwork) error {
+	delete(instance.Annotations, cloudManager.RestoreInProgress)
+	if !utils.ContainsString(instance.ObjectMeta.Finalizers, DataNetworkFinalizerName) {
+		instance.ObjectMeta.Finalizers = append(instance.ObjectMeta.Finalizers, DataNetworkFinalizerName)
+	}
+	err := r.Client.Update(context.TODO(), instance)
+	if err != nil {
+		return common.NewResourceStatusDependency(fmt.Sprintf("Failed to update '%s' data network resource after removing '%s' annotation during restoration.",
+			instance.Name, cloudManager.RestoreInProgress))
+	}
+	return nil
 }
--- a/controllers/host/host_controller.go
+++ b/controllers/host/host_controller.go
@@ -2110,6 +2110,22 @@ func (r *HostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (r
 	// Cancel any existing monitors
 	r.CloudManager.CancelMonitor(instance)
 
+	if r.checkRestoreInProgress(instance) {
+		r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Restoring '%s' host resource status without doing actual reconciliation", instance.Name)
+		if err := r.RestoreHostStatus(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		if err := r.ClearRestoreInProgress(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation &&
+		instance.Status.Reconciled {
+		return ctrl.Result{}, nil
+	}
+
 	if instance.DeletionTimestamp.IsZero() {
 		// Ensure that the object has a finalizer setup as a pre-delete hook so
 		// that we can delete any hosts that we have previously added.
@@ -2149,6 +2165,23 @@ func (r *HostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (r
 	profile, err := r.BuildAndValidateCompositeProfile(instance)
 	if err != nil {
 		return reconcile.Result{}, err
+	}
+
+	// Restore the host status
+	if r.checkRestoreInProgress(instance) {
+		r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Restoring '%s' host resource status without doing actual reconciliation", instance.Name)
+		if err := r.RestoreHostStatus(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		if err := r.ClearRestoreInProgress(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation &&
+		instance.Status.Reconciled {
+		return ctrl.Result{}, nil
 	}
 
 	// Update scope from configuration
@@ -2208,4 +2241,62 @@ func (r *HostReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&starlingxv1.Host{}).
 		Complete(r)
+}
+
+// Verify whether we have annotation restore-in-progress
+func (r *HostReconciler) checkRestoreInProgress(instance *starlingxv1.Host) bool {
+	restoreInProgress, ok := instance.Annotations[cloudManager.RestoreInProgress]
+	if ok && restoreInProgress != "" {
+		return true
+	}
+	return false
+}
+
+// Update status
+func (r *HostReconciler) RestoreHostStatus(instance *starlingxv1.Host) error {
+	annotation := instance.GetObjectMeta().GetAnnotations()
+	config, ok := annotation[cloudManager.RestoreInProgress]
+	if ok {
+		restoreStatus := &cloudManager.RestoreStatus{}
+		err := json.Unmarshal([]byte(config), &restoreStatus)
+		if err == nil {
+			if restoreStatus.InSync != nil {
+				instance.Status.InSync = *restoreStatus.InSync
+			}
+			instance.Status.Reconciled = true
+			instance.Status.ObservedGeneration = instance.ObjectMeta.Generation
+			instance.Status.DeploymentScope = "bootstrap"
+			instance.Status.StrategyRequired = "not_required"
+			err = r.Client.Status().Update(context.TODO(), instance)
+			if err != nil {
+				log_err_msg := fmt.Sprintf(
+					"Failed to update host status while restoring '%s' resource. Error: %s",
+					instance.Name,
+					err)
+				return common.NewResourceStatusDependency(log_err_msg)
+			} else {
+				StatusUpdate := fmt.Sprintf("Status updated for host resource '%s' during restore with following values: Reconciled=%t InSync=%t DeploymentScope=%s",
+					instance.Name, instance.Status.Reconciled, instance.Status.InSync, instance.Status.DeploymentScope)
+				r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, StatusUpdate)
+
+			}
+		} else {
+			r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Failed to unmarshal '%s'", err)
+		}
+	}
+	return nil
+}
+
+// Clear annotation RestoreInProgress
+func (r *HostReconciler) ClearRestoreInProgress(instance *starlingxv1.Host) error {
+	delete(instance.Annotations, cloudManager.RestoreInProgress)
+	if !utils.ContainsString(instance.ObjectMeta.Finalizers, HostFinalizerName) {
+		instance.ObjectMeta.Finalizers = append(instance.ObjectMeta.Finalizers, HostFinalizerName)
+	}
+	err := r.Client.Update(context.TODO(), instance)
+	if err != nil {
+		return common.NewResourceStatusDependency(fmt.Sprintf("Failed to update '%s' host resource after removing '%s' annotation during restoration.",
+			instance.Name, cloudManager.RestoreInProgress))
+	}
+	return nil
 }
--- a/controllers/manager/manager.go
+++ b/controllers/manager/manager.go
@@ -199,9 +199,10 @@ type PlatformManager struct {
 }
 
 type RestoreStatus struct {
-	InSync          *bool   `json:"inSync"`
-	Reconciled      *bool   `json:"reconciled"`
-	DeploymentScope *string `json:"deploymentScope"`
+	InSync           *bool   `json:"inSync"`
+	Reconciled       *bool   `json:"reconciled"`
+	DeploymentScope  *string `json:"deploymentScope"`
+	StrategyRequired *string `json:"strategyRequired"`
 }
 
 func NewStrategyStatus() *StrategyStatus {
--- a/controllers/manager/manager.go
+++ b/controllers/manager/manager.go
@@ -41,6 +41,7 @@ const (
 	// Defines annotation keys for resources.
 	NotificationCountKey = "deployment-manager/notifications"
 	ReconcileAfterInSync = "deployment-manager/reconcile-after-insync"
+	RestoreInProgress    = "deployment-manager/restore-in-progress"
 )
 
 const (
@@ -195,6 +196,12 @@ type PlatformManager struct {
 	vimClient                       *gophercloud.ServiceClient
 	PlatformNetworkReconcilerStatus bool
 	GetPlatformClientImpl           func(namespace string) *gophercloud.ServiceClient
+}
+
+type RestoreStatus struct {
+	InSync          *bool   `json:"inSync"`
+	Reconciled      *bool   `json:"reconciled"`
+	DeploymentScope *string `json:"deploymentScope"`
 }
 
 func NewStrategyStatus() *StrategyStatus {
--- a/controllers/manager/manager.go
+++ b/controllers/manager/manager.go
@@ -199,10 +199,7 @@ type PlatformManager struct {
 }
 
 type RestoreStatus struct {
-	InSync           *bool   `json:"inSync"`
-	Reconciled       *bool   `json:"reconciled"`
-	DeploymentScope  *string `json:"deploymentScope"`
-	StrategyRequired *string `json:"strategyRequired"`
+	InSync *bool `json:"inSync"`
 }
 
 func NewStrategyStatus() *StrategyStatus {
--- a/controllers/platformnetwork_controller.go
+++ b/controllers/platformnetwork_controller.go
@@ -1257,7 +1257,11 @@ func (r *PlatformNetworkReconciler) Reconcile(ctx context.Context, request ctrl.
 		if err := r.ClearAnnotation(instance); err != nil {
 			return reconcile.Result{}, err
 		}
-		return ctrl.Result{}, err
+		return ctrl.Result{}, nil
+	}
+
+	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation {
+		return ctrl.Result{}, nil
 	}
 
 	// Update scope from configuration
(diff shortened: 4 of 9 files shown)


controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1332,38 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+	return false
+}
+
+// Update status and clear annoatation
+func (r *PlatformNetworkReconciler) RestorePlatformNetworkStatus(instance *starlingxv1.PlatformNetwork) error {
+	r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Status without reconcilation: %t", instance.Status.InSync)
+	annotation := instance.GetObjectMeta().GetAnnotations()
+	if annotation != nil {
+		config, ok := annotation["kubectl.kubernetes.io/last-applied-configuration"]
```
sriram-gn, line 1350: This should use `cloudManager.RestoreInProgress` annotation key.

sriram-gn (changes requested):
Commit message should include test cases.

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1332,38 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		if ok {
+			platformNetworkStatus := &starlingxv1.PlatformNetworkStatus{}
+			err := json.Unmarshal([]byte(config), &platformNetworkStatus)
+			if err == nil {
+				instance.Status.InSync = platformNetworkStatus.InSync
+				instance.Status.Reconciled = platformNetworkStatus.Reconciled
+				instance.Status.DeploymentScope = platformNetworkStatus.DeploymentScope
+				err = r.Client.Status().Update(context.TODO(), instance)
+				if err != nil {
+					logPlatformNetwork.Error(err, "Failed to update platform network status")
```
sriram-gn, line 1360: Return err, not just log.
I think it would be safer to return one of those errors with `result = RetryTransientError` (eg. NewResourceStatusDependency)

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1332,38 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				}
+			}
+			// clear annotation RestoreInProgress
+			delete(instance.Annotations, cloudManager.RestoreInProgress)
```
sriram-gn, line 1364: Here, we are just deleting the annotation in the struct, need to update kubernetes resource mainly.
`r.Client.Update(context.TODO(), instance)`.
To handle update failure, I think it would be safer to return one of those errors with `result = RetryTransientError` (eg. NewResourceStatusDependency)

sriram-gn (changes requested):

controllers/platformnetwork_controller.go
```diff
@@ -1248,6 +1248,13 @@ func (r *PlatformNetworkReconciler) Reconcile(ctx context.Context, request ctrl.
 		return reconcile.Result{}, err
 	}
 
+	// Restore the PlatformNetwork status
+	if r.checkRestoreInProgress(instance) {
+		if err := r.RestorePlatformNetworkStatus(instance); err != nil {
+			return reconcile.Result{}, err
+		}
```
sriram-gn, line 1255: What if err==nil?
We would like to return err, whatever be the case, it could either be nil or be NewResourceStatusDependency error.

sriram-gn (changes requested):

sharbhar:
remove jira id
and add test plan.

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Changes are done

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Changes are done

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Returned with NewResourceStatusDependency

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Updated in kubernetes resource and handled failure

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1333,43 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	annotation := instance.GetObjectMeta().GetAnnotations()
+	if annotation != nil {
+		config, ok := annotation[cloudManager.RestoreInProgress]
+		if ok {
+			platformNetworkStatus := &starlingxv1.PlatformNetworkStatus{}
+			err := json.Unmarshal([]byte(config), &platformNetworkStatus)
+			if err == nil {
+				instance.Status.InSync = platformNetworkStatus.InSync
+				instance.Status.Reconciled = platformNetworkStatus.Reconciled
+				instance.Status.DeploymentScope = platformNetworkStatus.DeploymentScope
```
sriram-gn, line 1358: What if annotation doesn't provide correct values?
Suggest not to set values to `instance.Status.*` unless values are specified and correct.
Can we have following new struct in manager.go:
```
type RestoreStatus struct {
	InSync          *bool   `json:"inSync"`
	Reconciled      *bool   `json:"reconciled"`
	DeploymentScope *string `json:"deploymentScope"`
}
```

and use the struct `cloudManager.RestoreStatus` to unmarshal JSON across all the controllers?

1. Update `instance.Status.InSync` only when `RestoreStatus.InSync != nil`
2. Update  `instance.Status.Reconciled` only when `RestoreStatus.Reconciled != nil`
3. Update  `instance.Status.DeploymentScope` only when `RestoreStatus.DeploymentScope != nil && (strings.ToLower(*RestoreStatus.DeploymentScope) == "bootstrap" || strings.ToLower(*RestoreStatus.DeploymentScope) == "principal" )

I believe this would make the code more robust.

sriram-gn (changes requested):

controllers/platformnetwork_controller.go
```diff
@@ -1248,6 +1248,14 @@ func (r *PlatformNetworkReconciler) Reconcile(ctx context.Context, request ctrl.
 		return reconcile.Result{}, err
 	}
 
+	// Restore the PlatformNetwork status
+	if r.checkRestoreInProgress(instance) {
+		if err := r.RestorePlatformNetworkStatus(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, err
```
sriram-gn, line 1256: Should we rather return `ctrl.Result{}, nil` to be on a safer side, we really don't want to trigger reconciliation at any point after removing the restore-in-progress annotation unless it's coming from the user.
Either way, I'm expecting a test case to verify this.

sriram-gn (commented):
Please add test case where garbage values / bad json is set to `deployment-manager/restore-in-progress` annotation.
There should not be any other side-effects, other than status not being accurately updated which is acceptable.

There shouldn't be any reconciliation triggers especially.

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Added RestoreStatus struct and modified the code accordingly.

sriram-gn (commented):
Request another test case to verify Generation == ObservedGeneration, if not we should be updating `ObservedGeneration` as part of `RestorePlatformNetworkStatus`.
This applies to every controller.

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1333,51 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+// Update status and clear annoatation
+func (r *PlatformNetworkReconciler) RestorePlatformNetworkStatus(instance *starlingxv1.PlatformNetwork) error {
+	r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Status without reconcilation: %t", instance.Status.InSync)
+	annotation := instance.GetObjectMeta().GetAnnotations()
+	if annotation != nil {
+		config, ok := annotation[cloudManager.RestoreInProgress]
+		if ok {
+			restoreStatus := &cloudManager.RestoreStatus{}
+			err := json.Unmarshal([]byte(config), &restoreStatus)
+			if err == nil {
```
sriram-gn, line 1355: If json.Unmarshal returns err, log the error (logging should be enough, no need to return, typically it would be syntax errors).

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1333,51 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				if restoreStatus.InSync != nil {
+					instance.Status.InSync = *restoreStatus.InSync
+				}
+				if restoreStatus.Reconciled != nil {
+					instance.Status.Reconciled = *restoreStatus.Reconciled
+				}
+				if restoreStatus.DeploymentScope != nil &&
+					(strings.ToLower(*restoreStatus.DeploymentScope) == "bootstrap" ||
+						strings.ToLower(*restoreStatus.DeploymentScope) == "principal") {
```
sriram-gn, line 1364: What if restoreStatus.Reconciled is nil? The pointer would fail to resolve and panic.

Check to determine if restoreStatus.DeploymentScope is one of ("bootstrap", "principal") should be in nested conditional block.

sriram-gn (changes requested):

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1333,51 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	restoreInProgress, ok := instance.Annotations[cloudManager.RestoreInProgress]
+	if ok && restoreInProgress != "" {
+		return true
+	}
+	return false
+}
+
+// Update status and clear annoatation
+func (r *PlatformNetworkReconciler) RestorePlatformNetworkStatus(instance *starlingxv1.PlatformNetwork) error {
+	r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Status without reconcilation: %t", instance.Status.InSync)
```
sriram-gn, line 1348: Can we rephrase this:
`"Status updated for PlatformNetwork resource '%s' during restore with following values: Reconciled=%t InSync=%t DeploymentScope=%s", instance.Status.Reconciled, instance.Status.InSync, instance.Status.DeploymentScope`

and we can place this `r.ReconcilerEventLogger.NormalEvent` after `err = r.Client.Status().Update` when err == nil

sriram-gn (changes requested):

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Changes are done

controllers/platformnetwork_controller.go
MahalakshmiManthiram: logged the error

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Changes are done

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1333,59 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	if ok && restoreInProgress != "" {
+		return true
+	}
+	return false
+}
+
+// Update status and clear annoatation
+func (r *PlatformNetworkReconciler) RestorePlatformNetworkStatus(instance *starlingxv1.PlatformNetwork) error {
+	annotation := instance.GetObjectMeta().GetAnnotations()
+	if annotation != nil {
```
yjian118, line 1349: Already have it checked in line1252, no need to check again. Only need to restore the status in this function

MahalakshmiManthiram: Removed the check

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1334,56 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+// Verify whether we have annotation restore-in-progress
+func (r *PlatformNetworkReconciler) checkRestoreInProgress(instance *starlingxv1.PlatformNetwork) bool {
+	restoreInProgress, ok := instance.Annotations[cloudManager.RestoreInProgress]
+	if ok && restoreInProgress != "" {
+		return true
+	}
+	return false
+}
+
+// Update status and clear annoatation
```
hugonicodemos, line 1347: typo: annotation

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1334,56 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+func (r *PlatformNetworkReconciler) RestorePlatformNetworkStatus(instance *starlingxv1.PlatformNetwork) error {
```
hugonicodemos, line 1348: Please avoid creating nested functions. I recommend creating additional helper functions instead.

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1334,56 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			if err != nil {
+				return common.NewResourceStatusDependency(fmt.Sprintf("Failed to update platform network status while restoring '%s' platform network resource.", instance.Name))
+			} else {
+				r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Status updated for PlatformNetwork resource '%s' during restore with following values: Reconciled=%t InSync=%t DeploymentScope=%s",
+					instance.Name, instance.Status.Reconciled, instance.Status.InSync, instance.Status.DeploymentScope)
+			}
+		} else {
+			r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Failed to unmarshal '%s'", err)
+		}
+		// clear annotation RestoreInProgress
```
hugonicodemos, line 1380: I think is better to do this clean outside the function

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1334,56 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	return false
+}
+
+// Update status and clear annoatation
+func (r *PlatformNetworkReconciler) RestorePlatformNetworkStatus(instance *starlingxv1.PlatformNetwork) error {
+	annotation := instance.GetObjectMeta().GetAnnotations()
+	config, ok := annotation[cloudManager.RestoreInProgress]
+	if ok {
+		restoreStatus := &cloudManager.RestoreStatus{}
+		err := json.Unmarshal([]byte(config), &restoreStatus)
```
hugonicodemos, line 1353: check if there is an err first, and then keep the code normally:

if err {
   ....
}

...

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1334,56 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+					instance.Status.DeploymentScope = *restoreStatus.DeploymentScope
+				} else {
+					r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Invalid deploymentScope value '%s', scope will be set to boostrap", *restoreStatus.DeploymentScope)
+					instance.Status.DeploymentScope = "bootstrap"
+				}
+			}
+			err = r.Client.Status().Update(context.TODO(), instance)
+			if err != nil {
+				return common.NewResourceStatusDependency(fmt.Sprintf("Failed to update platform network status while restoring '%s' platform network resource.", instance.Name))
+			} else {
```
hugonicodemos, line 1373: else is not needed since the if statement already includes a return.

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1334,56 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+					strings.ToLower(*restoreStatus.DeploymentScope) == "principal" {
+					instance.Status.DeploymentScope = *restoreStatus.DeploymentScope
+				} else {
+					r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Invalid deploymentScope value '%s', scope will be set to boostrap", *restoreStatus.DeploymentScope)
+					instance.Status.DeploymentScope = "bootstrap"
+				}
+			}
+			err = r.Client.Status().Update(context.TODO(), instance)
+			if err != nil {
+				return common.NewResourceStatusDependency(fmt.Sprintf("Failed to update platform network status while restoring '%s' platform network resource.", instance.Name))
```
hugonicodemos, line 1372: try to create the err message before to avoid long lines

hugonicodemos (changes requested):
Thanks, just a few comments inline

controllers/platformnetwork_controller.go
```diff
@@ -1325,3 +1334,56 @@ func (r *PlatformNetworkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+	return false
+}
+
+// Update status and clear annoatation
+func (r *PlatformNetworkReconciler) RestorePlatformNetworkStatus(instance *starlingxv1.PlatformNetwork) error {
+	annotation := instance.GetObjectMeta().GetAnnotations()
+	config, ok := annotation[cloudManager.RestoreInProgress]
+	if ok {
+		restoreStatus := &cloudManager.RestoreStatus{}
```
yjian118, line 1352: Where's the update of the RestoreStatus?

controllers/manager/manager.go
```diff
@@ -197,6 +198,12 @@ type PlatformManager struct {
 	GetPlatformClientImpl           func(namespace string) *gophercloud.ServiceClient
 }
 
+type RestoreStatus struct {
```
yjian118, line 201: It's weird to have the RestoreStatus in the manager, but only designed for platform networks, can you rename it if you want to use it?

yjian118:
Looks like there's a bug in the latest patchset of the "RestoreStatus".

Further, need to update the tests with steps in details for understanding the tests passed

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Comments are addressed

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Typo corrected

controllers/platformnetwork_controller.go
MahalakshmiManthiram: If there is an error, it returns an error. If there is no error, we are just logging the successful status update with the values

controllers/manager/manager.go
MahalakshmiManthiram: RestoreStatus is not only designed for platform network.  It is common across all the controllers. It will be used by other controllers like ptp instance, ptp interface and system controller,

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Additional helper function for clearing annotation is created

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Changes are done

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Moved these cleaning code out of the function

controllers/platformnetwork_controller.go
```diff
@@ -1248,6 +1248,23 @@ func (r *PlatformNetworkReconciler) Reconcile(ctx context.Context, request ctrl.
...
+			return reconcile.Result{}, err
+		}
+		if err := r.ClearAnnotation(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation &&
+		instance.Status.Reconciled == true {
```
sriram-gn, line 1264: Just keep it as `instance.Status.Reconciled` instead of `instance.Stauts.Reconciled == true`.
This will also fix golang-lint issues.

controllers/ptpinstance_controller.go
```diff
@@ -582,6 +582,22 @@ func (r *PtpInstanceReconciler) Reconcile(ctx context.Context, request ctrl.Requ
...
+		if err := r.RestorePtpInstanceStatus(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		if err := r.ClearAnnotation(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation {
```
sriram-gn, line 597: Need to add `instance.Status.Reconciled ` just as in platformnetwork_controller.go.

```
		instance.Status.Reconciled {

```

controllers/ptpinterface_controller.go
```diff
@@ -606,6 +606,23 @@ func (r *PtpInterfaceReconciler) Reconcile(ctx context.Context, request ctrl.Req
...
+		if err := r.ClearAnnotation(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation &&
+		instance.Status.Reconciled == true {
+		return ctrl.Result{}, nil
+	}
```
sriram-gn, line 624: Just keep it as `instance.Status.Reconciled` instead of `instance.Stauts.Reconciled == true`.
This will also fix golang-lint issues.

controllers/system/system_controller.go
```diff
@@ -1548,6 +1550,22 @@ func (r *SystemReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+		if err := r.RestoreSystemStatus(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		if err := r.ClearAnnotation(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation {
```
sriram-gn, line 1565: Need to add `instance.Status.Reconciled ` just as in platformnetwork_controller.go.

```
		instance.Status.Reconciled {

```

sriram-gn (changes requested):
Updates missing for `host/host_controller.go` and `hostprofile_controller.go`

controllers/platformnetwork_controller.go
MahalakshmiManthiram: Changes are done

controllers/ptpinstance_controller.go
MahalakshmiManthiram: Changes are done

controllers/system/system_controller.go
MahalakshmiManthiram: Changes are done

controllers/ptpinterface_controller.go
MahalakshmiManthiram: Changes are done

controllers/ptpinterface_controller.go
```diff
@@ -674,3 +691,84 @@ func (r *PtpInterfaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+func (r *PtpInterfaceReconciler) checkRestoreInProgress(instance *starlingxv1.PtpInterface) bool {
+	restoreInProgress, ok := instance.Annotations[cloudManager.RestoreInProgress]
+	if ok && restoreInProgress != "" {
+		return true
+	}
+	return false
+}
+
+// Update status and clear annotation
+func (r *PtpInterfaceReconciler) RestorePtpinterfaceStatus(instance *starlingxv1.PtpInterface) error {
```
yjian118, line 705: I don't like such a long method spreading in all the controllers, can you refactor it to multiple methods to avoid the nested logical conditions?

controllers/ptpinterface_controller.go
```diff
@@ -674,3 +691,84 @@ func (r *PtpInterfaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	annotation := instance.GetObjectMeta().GetAnnotations()
+	config, ok := annotation[cloudManager.RestoreInProgress]
+	if ok {
+		restoreStatus := &cloudManager.RestoreStatus{}
+		err := json.Unmarshal([]byte(config), &restoreStatus)
+		if err == nil {
+			if restoreStatus.InSync != nil {
+				instance.Status.InSync = *restoreStatus.InSync
+			}
+			if restoreStatus.DeploymentScope != nil {
```
yjian118, line 715: Can we make the restore status and the final status to "bootstrap" only? It doesn't make sense to run DM in the background to update the configurations in this stage

controllers/ptpinterface_controller.go
```diff
@@ -674,3 +691,84 @@ func (r *PtpInterfaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+					instance.Status.DeploymentScope = *restoreStatus.DeploymentScope
+				} else {
+					r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Invalid deploymentScope value '%s', scope will be set to boostrap", *restoreStatus.DeploymentScope)
+					instance.Status.DeploymentScope = "bootstrap"
+				}
+			} else {
+				r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "The 'deploymentScope' is not specified, assuming 'bootstrap' scope.")
+				instance.Status.DeploymentScope = "bootstrap"
+			}
+			if restoreStatus.StrategyRequired != nil {
```
yjian118, line 727: Regarding the strategy required status, just update all of them to not_required

controllers/ptpinterface_controller.go
```diff
@@ -674,3 +691,84 @@ func (r *PtpInterfaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			}
+		} else {
+			r.ReconcilerEventLogger.NormalEvent(instance, common.ResourceUpdated, "Failed to unmarshal '%s'", err)
+		}
+	}
+	return nil
+}
+
+// Clear annotation RestoreInProgress
+func (r *PtpInterfaceReconciler) ClearAnnotation(instance *starlingxv1.PtpInterface) error {
```
yjian118, line 763: 1 Need to rename it to "ClearRestoreInProgress", it is not clearing all the annotations
2 Important: there are multiple clearAnnotation in all the controllers, do we need to wait for all the steps done before the actual clear annotation?

controllers/ptpinterface_controller.go
```diff
@@ -674,3 +691,84 @@ func (r *PtpInterfaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	delete(instance.Annotations, cloudManager.RestoreInProgress)
+	if !utils.ContainsString(instance.ObjectMeta.Finalizers, PtpInterfaceFinalizerName) {
```
yjian118, line 765: Why to add a finalizer here?

controllers/manager/manager.go
```diff
@@ -197,6 +198,12 @@ type PlatformManager struct {
...
+	InSync           *bool   `json:"inSync"`
+	DeploymentScope  *string `json:"deploymentScope"`
```
yjian118, line 203: I would suggest to update the scope to boostrap and required to not_required, mainly because we should not let DM to do anything during the restore period

yjian118 (commented):
Another question: do we need to update the host controller?

controllers/ptpinterface_controller.go
MahalakshmiManthiram: 1 -> acknowledged
2 -> restore-in-progress annotation is the first to be processed and hence the first to be cleared as well, no dependency on other annotations as such

controllers/ptpinterface_controller.go
MahalakshmiManthiram: In case of restore-in-progress annotation, it would set the deploymentScope, inSync, strategyRequired and reconciled status and end the reconciliation (no finalizer is added). We are adding it as part of restore-in-progress just in case user wants to delete an object and the DM has to handle it (in case of supported objects).

controllers/manager/manager.go
MahalakshmiManthiram: removed deploymentScope and strategyRequired from struct and backup shell script and had assigned them to "bootstrap" and "not_required" in DM

controllers/ptpinterface_controller.go
MahalakshmiManthiram: after removing deploymentScope and strategyRequired from struct the method has been cut short

controllers/ptpinterface_controller.go
MahalakshmiManthiram: Acknowledged

controllers/ptpinterface_controller.go
MahalakshmiManthiram: Changes are done

controllers/datanetwork_controller.go
```diff
@@ -524,6 +524,22 @@ func (r *DataNetworkReconciler) Reconcile(ctx context.Context, request ctrl.Requ
...
+		}
+		if err := r.ClearRestoreInProgress(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation {
+		return ctrl.Result{}, nil
+	}
```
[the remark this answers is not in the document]
MahalakshmiManthiram: Changes are done

controllers/host/host_controller.go
```diff
@@ -2151,6 +2151,23 @@ func (r *HostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (r
 		return reconcile.Result{}, err
 	}
 
+	// Restore the Ptp instance status
```
sriram-gn, line 2154: Correction -> "Restore the Host instance status".

controllers/host/host_controller.go
```diff
@@ -2151,6 +2151,23 @@ func (r *HostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (r
...
+			return reconcile.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation &&
+		instance.Status.Reconciled {
+		return ctrl.Result{}, nil
+	}
+
```
sriram-gn, line 2170: Need to move it around, right after fetching Host instance.

sriram-gn (changes requested):

controllers/host/host_controller.go
MahalakshmiManthiram: Changes are done

controllers/host/host_controller.go
MahalakshmiManthiram: Changes are done

sriram-gn (approved):

PR #353, opened 2024-04-26
Adds back route reconciliation on disabled host

yjian118:
This commit adds back the route reconciliation logic which was removed
in PR https://github.com/Wind-River/cloud-platform-deployment-manager/pull/316, ensures static routes available before the completion of
the initial unlock.

Test plan:
1. Deploy a DC with AIOSX subcloud.
2. Reconfigure the subcloud with an additional route, verifies the route
creation will not trigger the system config update strategy.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ COPY --from=builder /workspace/manager .
 ENTRYPOINT ["/manager"]
 
 # Copy the delve debugger into a debug image
-FROM ubuntu:mantic-20240216 as debug
+FROM ubuntu:mantic-20240405 as debug
 WORKDIR /
 RUN apt-get update && apt-get install -y tcpdump net-tools iputils-ping iproute2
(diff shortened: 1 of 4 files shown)


sriram-gn (commented):
I think we have to also update FindRouteUUID at this point:
https://github.com/Wind-River/cloud-platform-deployment-manager/blob/master/platform/info.go#L511

to normalize subnets before comparing.
This will be valid in case of IPv6 where `fcff:1:2:0:0::0` is same as `fcff:1:2::` but `strings.EqualFold` would return false if we compare them in their raw form.

yjian118:
> subnets before comparing.

As this is mainly a bug fix, can we make this improvement in a later commit?

hugonicodemos (approved):

sriram-gn:
> > subnets before comparing.
> 
> As this is mainly a bug fix, can we make this improvement in a later commit?

Agreed, LGTM.

sriram-gn (approved):

PR #354, opened 2024-04-30
SetSystemReady & SetSystemType During Restore

sriram-gn:
This commit sets SystemReady and SystemType during restore so that all other reconcilers can reconcile when subsequent DM configurations are applied post upgrade.

Test Cases:
1. PASS: Verify Day-1 configuration applied after successful restore gets reconciled.
2. PASS: Verify Day-2 configuration applied after successful restore gets reconciled.
3. PASS: Verify all other reconcilers are not notified due to change in state caused in SystemReconciler.

### DIFF
--- a/controllers/system/system_controller.go
+++ b/controllers/system/system_controller.go
@@ -1558,11 +1558,23 @@ func (r *SystemReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 
 		if platformClient == nil {
 			// Create the platform client without notifying other controllers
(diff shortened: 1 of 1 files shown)


hugonicodemos:
Can you create a test case for this?

sriram-gn:
> Can you create a test case for this?

Wish I could but we currently don't have supporting framework for host controller / system controller to write unit test cases to cover these changes.
Requesting if this could be covered in a separate commit focused on UT @hugonicodemos?

hugonicodemos (approved):
LGTM

yjian118 (approved):

PR #355, opened 2024-05-06
Support Platform Network Reconciliation - Day-1 Workflows

sriram-gn:
This commit enables all of Day-1 workflows related to platform network reconciliation.
Apart from that, there are certain mechanisms built in to handle notifying host controller whenever an instance of platform network resource and address pool resource is created or updated in a non-blocking way while taking care of potential issues due to concurrent operations.

Test Cases:
 1. PASS - Verify successful unlock of AIO-SX in Day-1 without platformnetwork / addresspool instances.
 2. PASS - Verify successful Day-2 operation on AIO-SX host without platformnetwork / addresspool instances.
 3. PASS - Verify reconfiguration is blocked for all networks in Day-1.
 4. PASS - Verify fresh configuration (even partial ie. single new stack) is successful for all networks other than oam / admin / mgmt.
 5. PASS - Verify fresh configuration (even partial) is blocked for networks such as oam / admin / mgmt.
 6. PASS - Verify IPv6 addresspool associated with pxeboot fails to reconcile without requeuing the request despite fresh configuration.
 7. PASS - Verify oam fails to reconcile if either 'floatingAddress' or 'gateway' is missing from the addresspool spec on AIO-SX.
 8. PASS - Verify mgmt / cluster-host / pxeboot fails to reconcile if any of 'floatingAddress', controller0Address' or 'controller1Address' are missing from the addresspool spec.
 9. PASS - Verify addresspool resource instances can be deleted with no impact on the system.
10. PASS - Verify platformnetwork resource instances can be deleted with no impact on the system.

### DIFF
--- a/common/utilities.go
+++ b/common/utilities.go
@@ -4,6 +4,8 @@
 package common
 
 import (
+	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/addresspools"
+	"github.com/gophercloud/gophercloud/starlingx/inventory/v1/networks"
 	"net"
 	"regexp"
 )
@@ -163,3 +165,72 @@ func DedupeSlice[T comparable](sliceList []T) []T {
     return list
 }
 */
+
+func NormalizeIPv6(address string) string {
+	var ret_ip string
+	ip := net.ParseIP(address)
+	if ip == nil {
+		// The address doesn't seem to be valid IPv6.
+		// Returning the address as is because semantic checks are
+		// beyond the scope of this function.
+		// Use it with along with "IsIPv6" function instead.
+		return address
+	}
+	ret_ip = ip.String()
+	return ret_ip
+}
+
+// IsIPAddressSame compares two addresses and returns
+// true when IP addresses are same and false when they
+// are not the same.
+func IsIPAddressSame(ip1 string, ip2 string) bool {
+	if ip1 != "" && IsIPv6(ip1) {
+		// Normalization of IP address is required in case of
+		// IPv6 due to different notations that could be employed.
+		// Example: fcff:1::0 is same as fcff:1:0:0:0::0 but string
+		// comparison would fail if IPs are not normalized.
+		ip1 = NormalizeIPv6(ip1)
+	}
+
+	if ip2 != "" && IsIPv6(ip2) {
+		ip2 = NormalizeIPv6(ip2)
+	}
+
+	return ip1 == ip2
+}
+
+func GetSystemAddrPoolByUUID(addrpool_list []addresspools.AddressPool, addrpool_uuid string) *addresspools.AddressPool {
+	for _, addrpool := range addrpool_list {
+		if addrpool.ID == addrpool_uuid {
+			return &addrpool
+		}
+	}
+	return nil
+}
+
+func GetSystemAddrPoolByName(addrpool_list []addresspools.AddressPool, addrpool_name string) *addresspools.AddressPool {
+	for _, addrpool := range addrpool_list {
+		if addrpool.Name == addrpool_name {
+			return &addrpool
+		}
+	}
+	return nil
+}
+
+func GetSystemNetworkByUUID(network_list []networks.Network, network_uuid string) *networks.Network {
+	for _, network := range network_list {
+		if network.UUID == network_uuid {
+			return &network
+		}
+	}
+	return nil
+}
+
+func GetSystemNetworkByName(network_list []networks.Network, network_name string) *networks.Network {
+	for _, network := range network_list {
+		if network.Name == network_name {
+			return &network
+		}
+	}
+	return nil
+}
--- a/controllers/addresspool_controller.go
+++ b/controllers/addresspool_controller.go
@@ -5,6 +5,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"github.com/go-logr/logr"
 	"github.com/gophercloud/gophercloud"
 	starlingxv1 "github.com/wind-river/cloud-platform-deployment-manager/api/v1"
@@ -13,7 +14,6 @@ import (
 	cloudManager "github.com/wind-river/cloud-platform-deployment-manager/controllers/manager"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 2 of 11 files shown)


controllers/addresspool_controller.go
```diff
@@ -13,7 +13,8 @@ import (
 	cloudManager "github.com/wind-river/cloud-platform-deployment-manager/controllers/manager"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
+	// "k8s.io/apimachinery/pkg/types"
```
yjian118, line 16: remove the package if no longer needed

controllers/manager/manager.go
```diff
@@ -593,6 +606,60 @@ func (m *PlatformManager) CancelMonitor(object client.Object) {
 	}
 }
 
+func (m *PlatformManager) GetActiveHostAndSystemInfo(namespace string, client *gophercloud.ServiceClient) (*v1.Host, *SystemInfo, error) {
```
yjian118, line 609: Suggest to split it into two methods: getActiveHost and getSystemInfo

controllers/host/host_controller.go
```diff
@@ -1486,29 +1482,29 @@ func (r *HostReconciler) StopAfterInSync() bool {
 
 // ReconcileExistingHost is responsible for dealing with the provisioning of an
 // existing host.
-func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient, instance *starlingxv1.Host, profile *starlingxv1.HostProfileSpec, host *hosts.Host) error {
+func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient, instance *starlingxv1.Host, profile *starlingxv1.HostProfileSpec, host *hosts.Host, request_namespace string) (error, []error) {
```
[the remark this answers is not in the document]
sriram-gn: That's a good point but I thought of having platformnetwork subreconciler's error captured separately so that it will be easier to implement specific exception handling for subreconciler's errors. Unlike other sub-reconcilers, platformnetwork sub-reconciler is not directly involved in host attributes reconciliation so there might be instances where we just want to let certain errors not block host reconciliation. What do you think @yjian118 ?

controllers/host/host_controller.go
```diff
@@ -1590,16 +1586,47 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
...
+
+			r.CloudManager.SetPlatformNetworkReconciling(false)
+			r.CloudManager.SetNotifyingActiveHost(false)
+
+			if len(platform_network_subreconciler_errs) != 0 {
+				return nil, platform_network_subreconciler_errs
+			}
+
+		} else {
+			err_msg := "platform network reconciliation is already in progress in active controller."
```
yjian118, line 1619: As it is expected to only update in the active controller, can we change the message to "waiting for platform network reconciled"?

controllers/platformnetwork_controller.go
```diff
@@ -385,7 +391,6 @@ func (r *PlatformNetworkReconciler) RestorePlatformNetworkStatus(instance *starl
 			instance.Status.Reconciled = true
 			instance.Status.ObservedGeneration = instance.ObjectMeta.Generation
 			instance.Status.DeploymentScope = "bootstrap"
-			instance.Status.StrategyRequired = "not_required"
```
yjian118, line 388: Why to remove the not_required here?

controllers/host/host_controller.go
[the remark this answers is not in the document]
yjian118: Not quite sure what do you mean "not block host reconciliation", as "inSync = (err == nil && platform_network_subreconciler_errs == nil)" is should be blocking the host status. Would suggest have clear message regarding the reason why it blocks. And you can define the type/struct of an error and guide the error handling by its type

controllers/platformnetwork_controller.go
sriram-gn: The "StrategyRequired" field is not part of PlatformNetworkStatus anymore.
It is also not part of AddressPoolStatus.
The reason is VIM checks "strategy_required" of Host resource instance only to propagate VIM strategies (ie. lock/unlock) and the field seemed to serve no purpose on PlatformNetwork / AddressPool resources.

common/utilities.go
```diff
@@ -163,3 +163,10 @@ func DedupeSlice[T comparable](sliceList []T) []T {
     return list
 }
 */
+
+func NormalizeIPv6(address string) string {
```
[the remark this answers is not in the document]
hugonicodemos: Can you handle the fail case?

    ip := net.ParseIP(address)
    if ip == nil {
       HANDLE FAIL
    }
    return ip.String()

controllers/addresspool_controller.go
```diff
@@ -62,20 +62,35 @@ func (r *AddressPoolReconciler) UpdateInsyncStatus(client *gophercloud.ServiceCl
...
-			logAddressPool.Error(err, "Failed to update status")
-			return err
+			host_instance.Status.InSync = false
+			host_instance.Status.Reconciled = false
+			err = r.Client.Status().Update(context.TODO(), host_instance)
+			if err != nil {
+				msg := fmt.Sprintf("Failed to update '%s' host instance status", host_instance.Name)
+				logAddressPool.Error(err, msg)
+				return common.NewResourceConfigurationDependency(msg)
+			} else {
```
hugonicodemos, line 79: dont need the else (return inside the if), please avoid to turn the methods more nested.

hugonicodemos, line 79: dont need the else (return inside the if), please avoid to turn the methods more nested.

controllers/host/host_controller.go
```diff
@@ -1590,9 +1586,45 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
...
+	}
+
+	system_info, err := r.CloudManager.GetSystemInfo(request_namespace, client)
+	if err != nil {
+		msg := "failed to get active host"
+		return common.NewUserDataError(msg)
+	}
+
+	if instance.Name == host_instance.Name {
+		// Only reconcile platform networks in active controller reconciliation
```
hugonicodemos, line 1602: Can we modularize this in a utility method?

controllers/host/platformnetworks.go
```diff
@@ -5,72 +5,953 @@ package host
...
-// and address pool instances to true just to acknowledge these objects are accepted.
-func (r *HostReconciler) ReconcilePlatformNetworks(client *gophercloud.ServiceClient, instance *starlingxv1.Host, profile *starlingxv1.HostProfileSpec, host *v1info.HostInfo) error {
-	if !utils.IsReconcilerEnabled(utils.HostPlatformNetwork) {
-		return nil
+// makeRangeArray converts an array of range structs to an array of arrays where
+// the inner array contains two elements.  The first element is the range start
+// address and the second element is the range end address.  This is to align
+// with the system API formatting which represents a pair as an array of two
+// elements.
+func makeRangeArray(ranges []starlingxv1.AllocationRange) [][]string {
```
hugonicodemos, line 29: these 2 method should be moved to a common utils file.

controllers/host/platformnetworks.go
```diff
@@ -5,72 +5,953 @@ package host
...
-		return err
+		logHost.Error(err, "failed to fetch addresspools from system")
+		return nil, err
+	}
+	// Preferably fetch the addresspool using UUID.
+	if addrpool_instance.Status.ID != nil {
+		found_addrpool = GetSystemAddrPoolByUUID(addrpool_list, *addrpool_instance.Status.ID)
+		if found_addrpool != nil {
+			return found_addrpool, nil
+		} else {
```
hugonicodemos, line 87: else not needed

hugonicodemos (changes requested):
Great job, @sriram-gn . Just few nits inline, and I recommend to move all these get by name/uuid functions to a common place, this could be necessary outside the platformnetwork controller and makes the file cleaner. thanks!

controllers/addresspool_controller.go
sriram-gn: Done

sriram-gn: Done

controllers/host/host_controller.go
sriram-gn: Done, modularized under `IsActiveHost` method in host controller.

controllers/host/platformnetworks.go
sriram-gn: Tried to move `makeRangeArray` and `compareRangeArrays` but then realized it causes circular import issue between `package v1` and `package common`, ie. `v1` imports `common` and to get `makeRangeArray` to work we have to import `v1` in `common` :)

controllers/host/platformnetworks.go
sriram-gn: Done

sriram-gn:
> Great job, @sriram-gn . Just few nits inline, and I recommend to move all these get by name/uuid functions to a common place, this could be necessary outside the platformnetwork controller and makes the file cleaner. thanks!

Thanks for reviewing @hugonicodemos!
I've moved name / uuid functions under utilities which definitely tidied platformnetworks.go a whole lot!

Also updated failure scenario in `NormalizeIPv6` and avoided `return` nested under `else` wherever I could.

It seems like we cannot move makeRangeArray under utilites and keeping compareRangeArrays following weird issue:
```
# github.com/wind-river/cloud-platform-deployment-manager/controllers/host
controllers/host/platformnetworks.go:256:39: undefined: utils.compareRangeArrays
# github.com/wind-river/cloud-platform-deployment-manager/controllers/host
vet: controllers/host/platformnetworks.go:256:39: undefined: utils.compareRangeArrays
```

So, I had to just let the two functions stick around in platformnetworks.go.

common/utilities.go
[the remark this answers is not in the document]
sriram-gn: Done

controllers/host/platformnetworks.go
```diff
@@ -5,72 +5,911 @@ package host
...
+	if addrpool_instance.Status.ID != nil {
+		found_addrpool = utils.GetSystemAddrPoolByUUID(addrpool_list, *addrpool_instance.Status.ID)
+		if found_addrpool != nil {
+			return found_addrpool, nil
+		}
+
+		// In case addresspool instance is using a stray ID fetch it by name.
+		found_addrpool = utils.GetSystemAddrPoolByName(addrpool_list, addrpool_instance.Name)
+
+	} else {
```
hugonicodemos, line 74: I don't think we need this else or the found_addrpool = utils.GetSystemAddrPoolByName(addrpool_list, addrpool_instance.Name) 2x. If found with UUID return, if not return using name.

controllers/host/platformnetworks.go
```diff
@@ -5,72 +5,911 @@ package host
...
+	if network_instance.Status.ID != nil {
+		found_network = utils.GetSystemNetworkByUUID(network_list, *network_instance.Status.ID)
+		if found_network != nil {
+			return found_network, nil
+		}
+
+		// In case network instance is using a stray ID fetch it by name.
+		found_network = utils.GetSystemNetworkByName(network_list, network_instance.Name)
+
+	} else {
```
hugonicodemos, line 99: ditto about else, and repetition

hugonicodemos (changes requested):

controllers/host/platformnetworks.go
sriram-gn: It's updated now, please check

hugonicodemos (approved):

controllers/platformnetwork_controller.go
yjian118: acked

yjian118 (approved):

PR #364, opened 2024-05-31
Support DeploymentScope Update When Generation equals ObservedGeneration

sriram-gn:
This commit fixes the bug where the user could not update deploymentScope without updating the spec of resource kinds such as Hosts, System, DataNetwork, PTPInterface and PTPInstance.

Test Cases:
1. PASS - Deploy AIO-SX system successfully with existing Day-1 configuration.
2. PASS - Verify that Day-2 operations are successful on AIO-SX.
3. PASS - Verify that deploymentScope of resource kinds such as Hosts, System, DataNetwork, PTPInterface and PTPInstance can be updated back and forth between "bootstrap" and "principal" without changin the spec.

### DIFF
--- a/controllers/datanetwork_controller.go
+++ b/controllers/datanetwork_controller.go
@@ -402,11 +402,13 @@ func (r *DataNetworkReconciler) GetScopeConfig(instance *starlingxv1.DataNetwork
 	return deploymentScope, nil
 }
 
-// Update deploymentScope and ReconcileAfterInSync in instance
+// Update ReconcileAfterInSync in instance
 // ReconcileAfterInSync value will be:
 // "true"  if deploymentScope is "principal" because it is day 2 operation (update configuration)
 // "false" if deploymentScope is "bootstrap"
 // Then reflect these values to cluster object
+// It is expected that instance.Status.Deployment scope is already updated by
+// UpdateDeploymentScope at this point.
 func (r *DataNetworkReconciler) UpdateConfigStatus(instance *starlingxv1.DataNetwork) (err error) {
 	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		err := r.Client.Get(context.TODO(), types.NamespacedName{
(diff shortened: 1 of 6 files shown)


controllers/datanetwork_controller.go
```diff
@@ -537,7 +537,14 @@ func (r *DataNetworkReconciler) Reconcile(ctx context.Context, request ctrl.Requ
...
+		// as ObservedGeneration, instance is reconciled.
+		err, scope_updated := r.UpdateDeploymentScope(instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		} else if !scope_updated {
+			return ctrl.Result{}, nil
+		}
 	}
 
 	// Update scope from configuration
```
yjian118, line 550: Expect to update the comment/log/method description around the UpdateConfigStatus. And if the update scope is done in UpdateDeploymentScope, it should not done again in UpdateConfigStatus

sriram-gn: Done, updated method description around UpdateConfigStatus across all controllers. Scope is either updated in UpdateDeploymentScope or UpdateConfigStatus (among other status parameters in this case) depending on Generation v/s ObservedGeneration comparison but will it be updated twice.

hugonicodemos (approved):

yjian118 (approved):

controllers/datanetwork_controller.go
yjian118: Doesn't make sense to have two methods to update scope twice, can you update the logic to avoid that?

sriram-gn: Done, separated "deploymentScope" update logic from "UpdateConfigStatus" function.

controllers/datanetwork_controller.go
```diff
@@ -536,8 +525,15 @@ func (r *DataNetworkReconciler) Reconcile(ctx context.Context, request ctrl.Requ
...
+	}
+
 	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation && instance.Status.Reconciled {
-		return ctrl.Result{}, nil
+		if !scope_updated {
+			return ctrl.Result{}, nil
+		}
 	}
 
 	// Update scope from configuration
```
yjian118, line 539: As deployment scope was no longer updated in UpdateConfigStatus(), expect to update the comment and long here

controllers/datanetwork_controller.go
```diff
@@ -407,6 +407,8 @@ func (r *DataNetworkReconciler) GetScopeConfig(instance *starlingxv1.DataNetwork
 // "true"  if deploymentScope is "principal" because it is day 2 operation (update configuration)
```
yjian118, line 407: Need to update the method description

sriram-gn: Done.

yjian118 (approved):

PR #365, opened 2024-05-31
Enhancement to avoid instant fatal errors

sselvara1:
AIO-SX Subcloud DM failure during scale deployment while evaluating the conditional check for get unreconciled status. This task is the 1st task after the host reboot after unlock, it may be a timing issue as some of the system service not was started at that time. Adding an enhancement to avoid instant fatal errors using proper "failed_when" to replace "ignore_errors: true". So the "retry" in this task can be respected.

Test Cases:
  1. PASS? Deploy 500 AIO-SX subclouds in parallel.

### DIFF
--- a/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
+++ b/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
@@ -712,7 +712,7 @@
               get_unrecon_status_post.stderr == "")
             retries: "{{ host_reconcile_retries }}"
             delay: 30
-            ignore_errors: yes
+            failed_when: false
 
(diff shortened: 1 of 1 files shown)


sriram-gn (commented):
Depending upon availability of resources and chances are sometimes deploying even 500 AIO-SX might not reveal this issue directly, given that,
would it be acceptable to test out `retries` are being respected with this update by using static columns like `.metadata.namespace` piped with `awk '$NF ~ /^deployment/ {print}`?

LGTM apart from concern related to testing.

hugonicodemos (approved):

sriram-gn (approved):
LGTM

PR #369, opened 2024-06-03
Fix IPv6 static mode support

enzocandotti:
This commit updates the hasIPv6StaticAddresses in order to fix the ipv6 static interface setup.

Test Plan:
PASS: Execute a Day1 operation creating a IPv6 static interface. Verify the interface is created correctly.


sriram-gn (approved):
Request if we could add Day-2 test case too.
Change LGTM, thanks Enzo!

hugonicodemos (approved):

PR #370, opened 2024-06-05
Support Network Creation with PoolUUID

sriram-gn:
It's not possible to create a network without PoolUUID as of now, hence, we have to support creation of network with "PoolUUID" which would mean reversal of reconciliation order from "network first" approach to "addresspool first" approach.

It's important to note that UUID of first addresspool in the list of "AssociatedAddressPools" will be used during creation of network and this will become the primary IP pool of the network.

Test Cases:
1. PASS - Verify storage network and addresspools are created successfully as per the specs applied.
2. PASS - Verify network of type "other" and associated addresspools are created successfully as per the specs applied.

### DIFF
--- a/controllers/host/platformnetworks.go
+++ b/controllers/host/platformnetworks.go
@@ -159,37 +159,42 @@ func (r *HostReconciler) ValidateAddressPool(
 
 // IsNetworkUpdateRequired determines if platform network update is required
 // by comparing it with applied PlatformNetwork spec.
-func (r *HostReconciler) IsNetworkUpdateRequired(network_instance *starlingxv1.PlatformNetwork, system_network *networks.Network) (opts networks.NetworkOpts, result bool, uuid string) {
+func (r *HostReconciler) IsNetworkUpdateRequired(network_instance *starlingxv1.PlatformNetwork, current_network *networks.Network, primary_address_pool *addresspools.AddressPool) (opts networks.NetworkOpts, result bool, uuid string) {
 	var delta strings.Builder
 
 	spec := network_instance.Spec
 
-	if system_network == nil || (network_instance.Name != system_network.Name) {
+	if current_network == nil || (network_instance.Name != current_network.Name) {
 		opts.Name = &network_instance.Name
 		delta.WriteString(fmt.Sprintf("\t+Name: %s\n", *opts.Name))
 		result = true
 	}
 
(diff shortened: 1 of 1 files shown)


yjian118:
"It's important to note that UUID of first addresspool in the list of "AssociatedAddressPools" will be used during creation of network and this will become the primary IP pool of the network."

Can we make it as a comment somewhere?

controllers/host/platformnetworks.go
```diff
@@ -184,6 +184,11 @@ func (r *HostReconciler) IsNetworkUpdateRequired(network_instance *starlingxv1.P
 
 	if system_network != nil {
 		uuid = system_network.UUID
+	} else {
```
yjian118, line 187: if system_network == nil {
    // Network is not configured in the system, assign the pool and create the network  
     new section
}

yjian118: BTW, it is confusing for me to think about what is "system_network", can you rename it as "current_network"?

sriram-gn: > if system_network == nil {
> // Network is not configured in the system, assign the pool and create the network
> new section
> }
Done.

> BTW, it is confusing for me to think about what is "system_network", can you rename it as "current_network"?
Done.

controllers/host/platformnetworks.go
```diff
@@ -184,6 +184,11 @@ func (r *HostReconciler) IsNetworkUpdateRequired(network_instance *starlingxv1.P
...
+		// This is required while creating the network only and
+		// internally managed by the system during reconfiguration.
+		opts.PoolUUID = &primary_address_pool.ID
+		delta.WriteString(fmt.Sprintf("\t+PoolUUID: %v\n", *opts.PoolUUID))
```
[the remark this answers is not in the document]
sriram-gn: Done.

sriram-gn:
> "It's important to note that UUID of first addresspool in the list of "AssociatedAddressPools" will be used during creation of network and this will become the primary IP pool of the network."
> 
> Can we make it as a comment somewhere?

Done, it's here:
https://github.com/Wind-River/cloud-platform-deployment-manager/pull/370/files#diff-5067c7a106f9179836e1f03276e0cdd0b67a3c437cec1ce2b0a80aed7986fbb7R881

yjian118 (approved):

PR #374, opened 2024-06-07
Support system config update strategy after factory install

yjian118:
Add support for factory install configuration
This commit adds support for configuration post factory install, this process is
similar to the day-2 operation, the differences:
1. The strategy is expected to be enforced once regardless if any configuration
expected the strategy to lock/unlock the hosts.
2. As some of the reconfigurations are expected to be done after the initial
deployment and before the reconfiguration, the host and system defaults values
are expected to be updated once to prevent unexpected rollbacks of the
configurations.
3. This configuration should not be triggered again once finalized.
4. During this process, deployment scope is not updated to principal as the
expectation is the deployment scope should be bootstrap.

With the request above, this commit give the DM the access of
configMaps, and leverage a factory install configMap to orchestate the
reconfiguration post factory install.

This commit also refactors some duplicated methods amongst the controllers and
adds additional unit test to cover the refactoring.

Test plan:
1. Passed - verified fresh deployment of simplex and duplex.
2. Passed - verified the reconfiguration after factory install w/o any
configuration change.
3. Passed - verified the reconfiguration after factory install with MTU
update.
4. Passed - verified the day-2 operation w/o factory install, updating fs size
does not trigger strategy.
5. Passed - verified the day-2 operation w/o factory install, updating MTC
tiggers strategy.
6. Passed - verified the day-2 operations above on the system after the
reconfiguration post factory install.
7. Passed - verified the DM update case.


yjian118:
Ready for review

DEBUGGING.md
```diff
@@ -6,14 +6,14 @@ This document provides information on the best approaches to debug DM issues on
 ## Checking deployment-manager status
 The following command will display the deployment-manager status:
 ```
-kubectl -n deployment get datanetworks,hostprofiles,hosts,platformnetworks,systems
+kubectl -n deployment get datanetworks,hostprofiles,hosts,platformnetworks,systems,ptpinstaces,ptpinterfaces
```
sriram-gn, line 9: Request to add "addresspools" as well.

DEBUGGING.md
```diff
@@ -6,14 +6,14 @@ This document provides information on the best approaches to debug DM issues on
...
 ```
 
 Of primary interest is the INSYNC column, which shows whether the deployment-manager
 has been able to synchronize the deployment configuration with the system, and
 the RECONCILED column, which shows that the synchronization completed once:
 ```
-controller-0:~$ kubectl -n deployment get datanetworks,hostprofiles,hosts,platformnetworks,systems
+controller-0:~$ kubectl -n deployment get datanetworks,hostprofiles,hosts,platformnetworks,systems,ptpinstaces,ptpinterfaces
```
sriram-gn, line 16: Request to add "addresspools" as well.

README.md
```diff
@@ -521,6 +521,35 @@ metadata:
...
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+	name: factory-install
+	namespace: deployment
+data:
+	factory-installed: "true"  # Initial state (change to "true" if factory-installed)
+	factory-config-finalized: "false" # Optional (change to "true" if configuration is complete)
+	system-abcd-default-updated: "false" # Optional (change to "true" if resource default updated)
```
sriram-gn, line 541: Does this key use following convention:
`<system-name>-default-updated`?

I think documentation related to this will be helpful.

README.md
```diff
@@ -521,6 +521,35 @@ metadata:
...
+  controller-0-default-updated: "false" # Optional (change to "true" if resource default updated)
```
sriram-gn, line 542: Does this key use following convention:
`<host-name>-default-updated`?

I think documentation related to this will be helpful.

controllers/common/update_scope_config.go
```diff
@@ -0,0 +1,101 @@
...
+		if ok {
+			scope, err := extractScope(config)
+			if err == nil {
+				if scope != "" {
+					scope, err = validateDeploymentScope(scope)
+					return scope, err
+				}
+			} else {
+				return scope, err
+			}
```
sriram-gn, line 62: Can we avoid nested `else` block and just return `scope, err` here?

controllers/manager/factory_install_info.go
```diff
@@ -0,0 +1,183 @@
...
+		// ConfigMap not found, return false with no error
+		log.V(2).Info("Factory install configmap not found", "namespace", ns)
+		return nil
+	}
+
+	// Update the factory-installed data field
+	if configMap.Data == nil {
+		configMap.Data = make(map[string]string)
+	}
+	configMap.Data["factory-config-finalized"] = strconv.FormatBool(value)
```
sriram-gn, line 104: Will it be better to use a const for `factory-config-finalized` key?

README.md
sriram-gn, line 541: Can we just use `system-default-update` instead?
I believe every DM deployment handles only one system.

README.md
sriram-gn, line 542: Indentation to be corrected?

controllers/manager/factory_install_info.go
```diff
@@ -0,0 +1,183 @@
...
+			// Any error other than NotFound is returned
+			return false, err
+		}
+		// ConfigMap not found, return false with no error
+		log.V(2).Info("Factory install configmap not found", "namespace", ns)
+		return false, nil
+	}
+
+	// Read the factory-installed status
+	factoryInstalledValue, exists := configMap.Data["factory-installed"]
```
sriram-gn, line 51: Will it be better to use a const for `factory-installed` key?

controllers/manager/factory_install_info.go
```diff
@@ -0,0 +1,183 @@
...
+	factoryInstalled, err := strconv.ParseBool(factoryInstalledValue)
+	if err != nil {
+		return false, err
+	}
+	if !factoryInstalled {
+		return false, err
+	}
+
+	// Read the factory-config-finalized status
+	factoryConfigFinalizedValue, exists := configMap.Data["factory-config-finalized"]
```
sriram-gn, line 66: Will it be better to use a const for `factory-config-finalized` key?

controllers/manager/factory_install_info.go
```diff
@@ -0,0 +1,183 @@
...
+	if !exists {
+		// installed but not finalized
+		return true, nil
+	}
+
+	factoryConfigFinalized, err2 := strconv.ParseBool(factoryConfigFinalizedValue)
+	if err2 != nil {
+		// installed but finalized value has error, expected to be updated later
+		return true, err
```
sriram-gn, line 75: Better to use `return factoryConfigFinalized, err`?

controllers/manager/factory_install_info.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	return !factoryConfigFinalized, nil
```
sriram-gn, line 78: If I've understood this correctly, should it be `return factoryConfigFinalized, nil` without the "NOT"?

sriram-gn (commented):
PR looks great with some awesome updates @yjian118!
Just posted some minor comments, feel free to resolve the conversation if some of these doesn't make sense!

sriram-gn (commented):
How do we plan to reset configMap values back to false in case of upgrade?

controllers/manager/factory_install_info.go
yjian118: No, if the factory config is finalized, it is expected to return false as the controllers should expect the factory install no longer expected

controllers/manager/factory_install_info.go
```diff
@@ -0,0 +1,183 @@
...
+	name: factory-install
+	namespace: deployment
+data:
+	factory-installed: "false"  # Initial state (change to "true" if factory-installed)
+	factory-config-finalized: "false" # Initial state (change to "true" if configuration is complete)
+	system-abcd-default-updated: "false" # Initial state (change to "true" if resource default updated)
+*/
+const FactoryInstallConfigMapName = "factory-install"
+
+func (m *PlatformManager) GetFactoryInstall(ns string) (bool, error) {
```
sriram-gn, line 32: Understood, does it make sense to rename the function to `IsFactoryInstallExpected`?
Or maybe a comment regarding the return values would be helpful.

README.md
yjian118: Would prefer to keep the method generic for all the resources, we need it for the systems and hosts currently though. The system name should expected to be correctly set, otherwise the DM is not good to handle multiple system resources in anyways.

README.md
yjian118: Yes, double checked

controllers/manager/factory_install_info.go
yjian118: make sense, I'm going to add a comment against this method

DEBUGGING.md
yjian118: Done

README.md
yjian118: Done

README.md
yjian118: Done

controllers/common/update_scope_config.go
yjian118: Re-structured the method, thanks for the advice

controllers/manager/factory_install_info.go
yjian118: Done

controllers/manager/factory_install_info.go
yjian118: Done

yjian118:
> How do we plan to reset configMap values back to false in case of upgrade?

In case of upgrade:
1 if the configmap is not restored, then the get factory install will return false, which is tested already that it will not impact the DM operation w/o the configmap(we will not have configmap in all the normal deployed system)
2 if the configmap is restored(mostly), the values will be persisted.
3 there's no expectation to have factory-installed=true but "factory-config-finalized=false" during upgrade, it will trigger the strategy and raise alarm to block the upgrade

sriram-gn (approved):
LGTM

hugonicodemos (approved):
LGTM

PR #386, opened 2024-07-12
Support Day-2 Platform Network Reconfig Workflows in Multinode Systems

sriram-gn:
Support Day-2 Platform Network Reconfig Workflows in Multinode Systems

In this commit we are enabling reconfiguration of platform networks such
as OAM and admin networks.

Reconfiguration of management network is blocked in multinode system and
OAM network reconfiguration involves lock and unlock of both the
controllers orchestrated by DM through VIM APIs.

The go.mod has also been updated to fix modification of range of the
addresspool during reconfiguration.

Test Cases:
1. PASS - Verify that range of the addresspool can be updated
          successfully for reconfigurable networks.
2. PASS - Verify that OAM, admin network reconfiguration is successful.
3. PASS - Verify that management network reconfiguration is blocked.
4. PASS - Verify that management network configuration is successful
          ie. the secondary stack.

### DIFF
--- a/controllers/host/host_controller.go
+++ b/controllers/host/host_controller.go
@@ -1,5 +1,5 @@
 /* SPDX-License-Identifier: Apache-2.0 */
-/* Copyright(c) 2019-2023 Wind River Systems, Inc. */
+/* Copyright(c) 2019-2024 Wind River Systems, Inc. */
 
 package host
 
@@ -140,7 +140,7 @@ const (
 )
 
 func (r *HostReconciler) IsActiveHost(client *gophercloud.ServiceClient, instance *starlingxv1.Host, reqNs string) (bool, error) {
-	host_instance, err := r.CloudManager.GetActiveHost(reqNs, client)
+	host_instance, _, err := r.CloudManager.GetHostByPersonality(reqNs, client, cloudManager.ActiveController)
 	if err != nil {
 		msg := "failed to get active host"
 		return false, common.NewUserDataError(msg)
@@ -667,7 +667,8 @@ func (r *HostReconciler) ReconcileInitialState(client *gophercloud.ServiceClient
 	desiredState := profile.AdministrativeState
 
 	if desiredState != nil && *desiredState != host.AdministrativeState &&
-		instance.Status.DeploymentScope == cloudManager.ScopeBootstrap {
+		instance.Status.DeploymentScope == cloudManager.ScopeBootstrap &&
+		!r.CloudManager.GetStrategySent() {
 		if *desiredState == hosts.AdminLocked {
 			action := hosts.ActionLock
 			opts := hosts.HostOpts{
@@ -1118,7 +1119,7 @@ func (r *HostReconciler) ReconcileHostByState(
 		// Update strategy regardless the disabled attributes
 		// the admin network mult-netting can also be done when host is disabled
 		// StrategyRequired should not be updated when strategy was sent.
-		if factory && !r.CloudManager.GetStrageySent() {
+		if factory && !r.CloudManager.GetStrategySent() {
 			logHost.Info("set lock required for configuration after factory install")
 			instance.Status.StrategyRequired = cloudManager.StrategyLockRequired
 			r.CloudManager.SetResourceInfo(cloudManager.ResourceHost, host.Personality, instance.Name, instance.Status.Reconciled, instance.Status.StrategyRequired)
@@ -1173,7 +1174,8 @@ func (r *HostReconciler) ReconcileHostByState(
 		}
 
 		// Clean up strategy required after Disabled and enabled attributes are all in-sync
-		if strategy_required {
+		if strategy_required &&
+			!(r.CloudManager.GetStrategyExpectedByOtherReconcilers() || r.CloudManager.GetStrategySent()) {
 			logHost.V(2).Info("set strategy not required as attributes are all configured")
 			instance.Status.StrategyRequired = cloudManager.StrategyNotRequired
 			r.CloudManager.SetResourceInfo(cloudManager.ResourceHost, host.Personality, instance.Name, instance.Status.Reconciled, instance.Status.StrategyRequired)
@@ -1387,7 +1389,8 @@ func (r *HostReconciler) statusUpdateRequired(instance *starlingxv1.Host, host *
 			logHost.V(2).Info("host inSync, set unlock required")
 			status.StrategyRequired = cloudManager.StrategyUnlockRequired
 			strategyUpdated = true
-		} else if status.StrategyRequired != cloudManager.StrategyNotRequired {
+		} else if status.StrategyRequired != cloudManager.StrategyNotRequired &&
+			!(r.CloudManager.GetStrategyExpectedByOtherReconcilers() || r.CloudManager.GetStrategySent()) {
 			logHost.V(2).Info("set not required: reconcile finished")
 			status.StrategyRequired = cloudManager.StrategyNotRequired
 			strategyUpdated = true
@@ -2235,6 +2238,7 @@ func (r *HostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (r
 	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation &&
 		instance.Status.Reconciled &&
 		instance.Status.DeploymentScope == "bootstrap" &&
+		instance.Status.StrategyRequired == cloudManager.StrategyNotRequired &&
 		!platformnetwork_update_required {
 
 		factory, err := r.GetFactoryInstall(request.Namespace)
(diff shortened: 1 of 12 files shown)


controllers/host/host_controller.go
```diff
@@ -1173,7 +1173,7 @@ func (r *HostReconciler) ReconcileHostByState(
 		}
 
 		// Clean up strategy required after Disabled and enabled attributes are all in-sync
-		if strategy_required {
+		if instance.Status.StrategyRequired == cloudManager.StrategyUnlockRequired && !r.CloudManager.GetStrategySent() {
```
yjian118, line 1176: Any reason to check if the strategy was sent and why the "LockRequired" was moved out of the scope?
Worrying about the cases, e.g.:
Strategy failed to be built, and the host was lock required. If we want to change the configuration back to what was previously insync, will the strategy lock required be stuck in this state?

Need also update the comment above to explain why we are not clearing up the strategy status even in-sync

controllers/host/host_controller.go
```diff
@@ -1387,7 +1387,7 @@ func (r *HostReconciler) statusUpdateRequired(instance *starlingxv1.Host, host *
 			logHost.V(2).Info("host inSync, set unlock required")
 			status.StrategyRequired = cloudManager.StrategyUnlockRequired
 			strategyUpdated = true
-		} else if status.StrategyRequired != cloudManager.StrategyNotRequired {
+		} else if status.StrategyRequired == cloudManager.StrategyUnlockRequired && host.IsUnlockedEnabled() {
```
yjian118, line 1390: Same comment here:
what's the expectation:
1 if strategy lock required
2 what if host in other status, failed/disabled etc.
When adding condition, please consider other conditions that cause code escape

controllers/host/platformnetworks.go
```diff
@@ -595,6 +595,15 @@ func (r *HostReconciler) ReconcilePlatformNetworkExpected(client *gophercloud.Se
 		case cloudManager.OAMNetworkType,
 			cloudManager.MgmtNetworkType,
 			cloudManager.AdminNetworkType:
+
+			// Reconciliation is not expected for management network on
+			// AIO-DX / Standard setups.
+			if system_info.SystemMode != cloudManager.SystemModeSimplex &&
```
yjian118, line 601: Why do we want to check the system mode? Isn't it expected to be rejected by the sysinv API?

yjian118:
Can you clarify the statement: "reconfiguration involves lock and unlock of both the controllers orchestrated by DM through VIM APIs."?
DM should not be responsible for orchestrating the hosts operations, we need avoid the steps managed by two entities

controllers/manager/monitor.go
```diff
@@ -429,6 +445,27 @@ func ManageStrategy(management CloudManager) bool {
 			_, err := management.GcCreate(client, request)
 			if err != nil {
 				log.Error(err, "Strategy creation failed")
+				// If existing strategy is in applied state delete it so that
+				// new strategy can be created, otherwise, monitor ongoing
+				// strategy.
+				if IsExistingStrategy(err) {
+					strategy_status, _ := management.GcShow(client)
```
yjian118, line 452: Can we move the GcShow above to avoid the GcCreate?

controllers/host/host_controller.go
sriram-gn: Reverted above condition with slight improvement so that it can handle corner cases where strategy is in progress or strategy creation request yet to be sent.

controllers/host/host_controller.go
sriram-gn: Reverted this condition with same improvement mentioned above.

controllers/manager/monitor.go
sriram-gn: Done.

controllers/host/platformnetworks.go
```diff
@@ -1110,6 +1130,79 @@ func (r *HostReconciler) ReconcilePlatformNetworkPrincipalSimplex(
...
+		return err
+	}
+
+	_, update_required, _ := r.IsAddrPoolUpdateRequired(network_instance, addrpool_instance, system_addrpool)
+
+	validation_result := r.ValidateAddressPool(network_instance, addrpool_instance, system_info)
+
+	if update_required && validation_result {
+
+		active_controller_instance, active_controller_obj, err := r.CloudManager.GetHostByPersonality(host_instance.Namespace, client, cloudManager.ActiveController)
```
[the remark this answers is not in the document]
sriram-gn: We have to set both active and standby controller's StrategyRequired to "lock_required" so that VIM takes that into account during strategy build phase.
During OAM reconfiguration lock and unlock is required on both the controllers.

sriram-gn:
> Can you clarify the statement: "reconfiguration involves lock and unlock of both the controllers orchestrated by DM through VIM APIs."? DM should not be responsible for orchestrating the hosts operations, we need avoid the steps managed by two entities


Orchestrated meaning DM just sets "lock_required" and everything else is handled by VIM, no more orchestrations from DM side, it would just be monitoring the strategy progress and clean it up in the end.

controllers/host/platformnetworks.go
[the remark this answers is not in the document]
yjian118: That's the reason I don't like to differ the "active controller" and "standby controller" if the operation is expected only on "controllers" rather than other personalities, can you add comments on that?

controllers/manager/monitor.go
```diff
@@ -425,8 +425,40 @@ func ManageStrategy(management CloudManager) bool {
 			log.Info("Vim client is not ready. Wait")
 			return false
 		} else {
+
+			strategy_status, err := management.GcShow(client)
```
yjian118, line 429: Can we move this section to a new method and give proper comment? Hard to read this block tbh?

controllers/host/host_controller.go
```diff
@@ -2470,21 +2473,23 @@ func (r *HostReconciler) GetAddressPoolsFromPlatformNetwork(associated_addrpools
...
 
-// LockHostRequest takes the host object, host ID / personality and reconcile_object
-// which describes the purpose of requesting locking of particular host, starts lock
-// host monitor and returns LockedDisabledHost monitor error.
+// LockHostRequest takes the host object, host ID / personality and
+// starts lock host monitor and returns LockedDisabledHost monitor error.
 // This would initiate lock request of a host through VIM so that disabled attributes
 // can be reconciled later.
-func (r *HostReconciler) LockHostRequest(host_instance *starlingxv1.Host, host_id string, host_personality string, reconcile_object string) error {
+func (r *HostReconciler) LockHostRequest(host_instance *starlingxv1.Host, host_id string, host_personality string, set_res_info bool) error {
```
yjian118, line 2480: I only see the SetWaitForStrategySent(true) in this method, but the LockHostReqeust() doesn't cover all the reconfiguration path

sriram-gn: In case of oam network reconfiguration the procedure is:
1. Reconfigure oam
2. Lock and unlock controllers
So, the reconfiguration is already done prior to lock / unlock in this case.

hugonicodemos (commented):
LGTM, but I'll approve only after a green sanity. Sounds good?

controllers/host/host_controller.go
```diff
@@ -1173,7 +1173,8 @@ func (r *HostReconciler) ReconcileHostByState(
 		}
 
 		// Clean up strategy required after Disabled and enabled attributes are all in-sync
-		if strategy_required {
+		if strategy_required &&
+			!(r.CloudManager.IsWaitingForStrategySent() || r.CloudManager.GetStrategySent()) {
```
yjian118, line 1177: I think the major problem is this condition: you want to prevent the strategy_required to be overridden to not required too early, which may happen if the strategy was not created yet.

I'm okay with this conditional check. Just expect to enhance the "IsWaitForStrategysent()":

- it actually does not check the strategy status, can we rename  as something like e.g. "GetStrategyExpectedByOtherReoncilers"
- Add debug/v2 level message to indicate the check of the prop
- Update the description above(better have a decision tree) to help to read

- 

1.

controllers/host/host_controller.go
yjian118: Can we have it as "LockHostRequestByOtherController/reconciler"? We are not using it in the host reconciler, right?
Can we have an info message to indicate the sent and update operation? note: we don't want it repeatedly dump into the logs, expected to have some prechecks

controllers/host/host_controller.go
sriram-gn: Done

controllers/host/host_controller.go
sriram-gn: Done

controllers/manager/monitor.go
sriram-gn: Done

sriram-gn:
> LGTM, but I'll approve only after a green sanity. Sounds good?

Sounds good, thanks @hugonicodemos

controllers/host/platformnetworks.go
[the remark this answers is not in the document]
sriram-gn: Done

hugonicodemos (approved):

controllers/manager/monitor.go
```diff
@@ -425,8 +474,27 @@ func ManageStrategy(management CloudManager) bool {
...
+			}
+
+			if is_existing_strategy && IsFinishedStrategy(strategy_status) {
+				err = DeleteFinishedStrategy(management, client)
+				if err != nil {
+					return false
+				}
+
+			} else if is_existing_strategy {
+				_ = MonitorExistingStrategy(strategy_status, management)
```
yjian118, line 490: The error  of this method is not tracked, if we don't want to track the result, suggest to remove the return value, otherwise need to consider the workflow

controllers/host/host_controller.go
```diff
@@ -667,7 +667,8 @@ func (r *HostReconciler) ReconcileInitialState(client *gophercloud.ServiceClient
 	desiredState := profile.AdministrativeState
 
 	if desiredState != nil && *desiredState != host.AdministrativeState &&
-		instance.Status.DeploymentScope == cloudManager.ScopeBootstrap {
+		instance.Status.DeploymentScope == cloudManager.ScopeBootstrap &&
+		!r.CloudManager.GetStrategySent() {
```
yjian118, line 671: Why do we want to check the strategy here? Isn't the check of the scope sufficient?

sriram-gn: No, in case of network reconfiguration, scope is not changed for host, it will only be updated for platformnetwork so it'll not help.

controllers/host/host_controller.go
```diff
@@ -2235,6 +2238,7 @@ func (r *HostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (r
 	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation &&
 		instance.Status.Reconciled &&
 		instance.Status.DeploymentScope == "bootstrap" &&
+		instance.Status.StrategyRequired == cloudManager.StrategyNotRequired &&
```
yjian118, line 2241: Same comment here: if the scope is "bootstrap", why do we want to check the strategy again?

sriram-gn: Same answer applies here as well:

> No, in case of network reconfiguration, scope is not changed for host, it will only be updated for platformnetwork so it'll not help.

controllers/manager/monitor.go
sriram-gn: Done.

controllers/host/host_controller.go
yjian118: acked

controllers/manager/monitor.go
```diff
@@ -341,6 +341,53 @@ func monitorStrategyState(management CloudManager) bool {
 	return false
 }
 
+func MonitorExistingStrategy(strategy_status *systemconfigupdate.SystemConfigUpdate, management CloudManager) {
+	log.Info(fmt.Sprintf(
```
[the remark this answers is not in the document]
sriram-gn: Done.

PR #387, opened 2024-07-22
DM playbook does not allow custom DM app path as input

vineela-pachchipulusu:
DM pods are set up on systems by the DM playbook by applying the DM app tarball, which is provided on /usr/local/share/applications/helm/ on WRCP ISOs.

However, the app tarball path variable used on the playbook is not configurable. For instance, if the dev moves the tarball from the original location and sets the 'manager_app_chart' variable in the overrides, the playbook will fail.

This isn't a major issue on WRCP ISOs, but it makes the playbook unusable on STX.

Test Plan:
Build wrcp-deployment-manager, all the dm packages.
AIO-SX deployment.


sharbhar:
LGTM

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -27,7 +27,12 @@
       when: deployment_config is defined
 
     - set_fact:
-        manager_app_chart: "{{ manager_app_chart | default('/usr/local/share/applications/helm/deployment-manager-*.tgz') }}"
+        manager_app_chart: "{{ manager_app_chart }}"
```
yjian118, line 30: Those two set_fact tasks are useless unless you want to set default values

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -197,7 +202,7 @@
       block:
       - name: Find deployment-manager app helm-chart in application directory
         find:
-          paths: /usr/local/share/applications/helm/
+          paths: application_helm_chart_dir
```
yjian118, line 205: If defaults are not set, this task will fail due to not defined

yjian118 (changes requested):
Expect default values to prevent overrides not offered

sselvara1 (approved):
LGTM

yjian118 (changes requested):
Need to at least test the aiosx deployment with this PR

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -27,7 +27,12 @@
       when: deployment_config is defined
 
     - set_fact:
-        manager_app_chart: "{{ manager_app_chart | default('/usr/local/share/applications/helm/deployment-manager-*.tgz') }}"
+        manager_app_chart: "{{ manager_app_chart | default('wind-river-cloud-platform-deployment-manager.tgz') }}"
```
yjian118, line 30: After changing the default value here, is the 'wind-river-cloud-platform-deployment-manager.tgz' able to be uploaded?

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
vineela-pachchipulusu: Have Set the default values

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
vineela-pachchipulusu: Handled

vineela-pachchipulusu:
> Need to at least test the aiosx deployment with this PR

Have updated with the default values and the AIO-SX VM is tested

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
vineela-pachchipulusu: Updated with the proper default value

hugonicodemos (approved):

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -197,7 +202,7 @@
       block:
       - name: Find deployment-manager app helm-chart in application directory
         find:
-          paths: /usr/local/share/applications/helm/
+          paths: application_helm_chart_dir
           patterns: 'deployment-manager*.tgz'
         register: app_helm_chart
```
yjian118, line 207: What's the app_helm_chart usage here if it was uploaded as manager_app_chart in line215?

yjian118:
1. Please test the case the tarball was not in the default directory(in sysadmin home), and add the playbook format and overrides content in the commit message.
2. Request to update the read me to reflect this update

yjian118:
Expect test case update:

Transffered the dm_app_helm_chart from the /usr/local/share/applications/helm/ to the /home/sysadmin/ path
Delete dm_app_helm_chart from the /usr/local/share/applications/helm/ to the /home/sysadmin/ path

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -17,17 +17,18 @@
...
         deploy_config: "{{ deployment_config }}"
         wait_for_dm_unlock: " {{ wait_for_dm_unlock| default(5) }}"
       when: deployment_config is defined
 
     - set_fact:
         manager_app_chart: "{{ manager_app_chart | default('/usr/local/share/applications/helm/deployment-manager-*.tgz') }}"
+      when: manager_app_chart is defined
+
+    - set_fact:
+        application_helm_chart_dir: "{{ application_helm_chart_dir | default('/usr/local/share/applications/helm/') }}"
```
yjian118, line 30: TBH, I don't know the purpose why we need an additional variable "application_helm_chart_dir", we can simply override the variable manager_app_chart. 
And it is dangerous to find a patterns: 'deployment-manager*.tgz', there may be more than one file has this pattern. And the register app_helm_chart was never used

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -17,17 +17,14 @@
...
-      when: deployment_manager_overrides is defined
-
     - set_fact:
         deploy_config: "{{ deployment_config }}"
         wait_for_dm_unlock: " {{ wait_for_dm_unlock| default(5) }}"
       when: deployment_config is defined
 
     - set_fact:
         manager_app_chart: "{{ manager_app_chart | default('/usr/local/share/applications/helm/deployment-manager-*.tgz') }}"
+      when: manager_app_chart is defined
```
yjian118, line 27: This condition should be removed, we want to use the default if not defined

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -17,17 +17,14 @@
         wait_for_timeout: " {{ wait_for_timeout | default(400) }}"
         boot_wait_time: " {{ boot_wait_time | default(180) }}"
 
-    - set_fact:
```
yjian118, line 20: Suggest to keep this task, otherwise the later tasks may fail.

Further, expect to change the task in line102 and 103 to avoid using hard-coded values, instead, set the hard-coded path as the default values, so we can have them as overrides

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -103,7 +103,7 @@
 
         - set_fact:
             helm_chart_overrides: "/usr/local/share/applications/overrides/wind-river-cloud-platform-deployment-manager-overrides-subcloud.yaml"
-            manager_chart: "/usr/local/share/applications/helm/wind-river-cloud-platform-deployment-manager-*.tgz"
+            manager_chart: {{ manager_chart }}
```
yjian118, line 106: This set fact is not necessary only if you want to set default value, but it was already set, let remove this line

yjian118 (approved):

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -197,7 +197,7 @@
       block:
       - name: Find deployment-manager app helm-chart in application directory
         find:
-          paths: /usr/local/share/applications/helm/
+          paths: "{{ manager_app_chart | dirname }}"
           patterns: 'deployment-manager*.tgz'
```
yjian118, line 201: I'm fine with this change for now as the expectation of the tarball name is like this pattern. But keep in mind, if we rename the tarball name and give it to the overrides, this pattern may not fit therefore the failure will be caused by it.

In these cases, we may need to think tasks to check if there's a wildcard in the file name, and process it differently

yjian118 (approved):

PR #391, opened 2024-08-15
Limit resource consumption

yjian118:
This commit limits the resource comsumption.

Further, due to the system configuration collection may impact the
sysinv-api heavily, this commit raise the threshhold of the DM pod
restart caused by the liveness probe and readiness probe.

This commit also disabled the leader election if the replica is less
than 2.

Test plan:
1. Deploy AIOSX/DX subcloud, ensure the playbook not failed due to the
slow down of the limited resource.
2. Trigger day-2 operation on the SX and DX subcloud, ensure the Day-2
operation not failed in this limited resource configuration.
3. Soak the DM in a special 1 platform core system, monitor the DM pod
no longer restarted frequently due to probe failure.


enzocandotti (approved):
LGTM based on test plan

yjian118:
> LGTM, but can we do the leader-elect check inside the code instead of commenting/uncommenting? Also, could you check why the image failed? thanks

Good advice, updated, the leader election is now configured based on the replica of the pod. The pod failure was caused by probe failure(timeout) in a resource-limited system, by extend the error threshold of the probe, the pod was not observed a restart after this commit in 48 hrs

PR #397, opened 2024-09-02
Restrict Filesystem Removal

hedibertocsilva:
Although the ceph host filesystem is optional and allowed to be created, its removal is only supported in specific situations and should not be addressed by the Deployment Manager.

This commit also resolves a reconcile error due to the automatic creation of the ceph host filesystem using Ceph bare metal.

Includes:
- Creates the FileSystemDeletionAllowed list excluding ceph.
- Updates the "Should return false when in and other storage is not equal and have allowed filesystem types" feature test to use the image-conversion instead of ceph.
- Creates a new feature test expecting the reconciliation to be true when the extra filesystem is ceph.

### DIFF
--- a/controllers/host/host_controller.go
+++ b/controllers/host/host_controller.go
@@ -70,6 +70,7 @@ var CephPrimaryGroup []string
 
 // Only the listed file systems are allow to create and delete
 var FileSystemCreationAllowed = []string{"instances", "image-conversion", "ceph"}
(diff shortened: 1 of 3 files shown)


hugonicodemos (approved):
LGTM

hugonicodemos:
Can you update the description with the Test Plan section? @hedibertocsilva

yjian118 (approved):

yjian118 (approved):

PR #398, opened 2024-09-04
Enable playbook task for ignoring unreachable servers

sselvara1:
DM playbook failed due to second reboot on standard subcloud.
    Playbook has block for when system reboot twice but it was not
    getting triggered. When server is unreachable, task status is
    set to unreachable instead of failed. From ansible 2.7, it is
    possible to use ignore_unreachable.

    Test Cases:
    1. Pending: Run dcmanager add from system controller. Deployment
       config is applied successfully on controller-0 with server
       rebooting twice. 'deploy status' shall be 'complete' for the
       subcloud.


sselvara1:
incorrect rebase.

PR #400, opened 2024-09-04
kubectl delete host resource failed by cmd timeout

vineela-pachchipulusu:
The kubectl delete host is getting stuck and failed by cmd timeout till the finalizer is manually deleted in the STANDARD setup and during the deletion of the compute host

Test Cases:
PASS- kubectl delete host should not timeout

### DIFF
--- a/controllers/host/host_controller.go
+++ b/controllers/host/host_controller.go
@@ -1842,6 +1842,14 @@ func (r *HostReconciler) ReconcileDeletedHost(client *gophercloud.ServiceClient,
 	return nil
 }
 
+// removing the HostFinalizer from the host resource
+func (r *HostReconciler) removeHostFinalizer(instance *starlingxv1.Host) {
+	instance.ObjectMeta.Finalizers = utils.RemoveString(instance.ObjectMeta.Finalizers, HostFinalizerName)
+	if err := r.Client.Update(context.Background(), instance); err != nil {
+		logHost.Error(err, "failed to remove the finalizer in the host because of the error:%v")
+	}
+}
+
 // ReconcileResource interacts with the system API in order to reconcile the
 // state of a data network with the state stored in the k8s database.
 func (r *HostReconciler) ReconcileResource(
@@ -1875,6 +1883,11 @@ func (r *HostReconciler) ReconcileResource(
 
 	if !instance.DeletionTimestamp.IsZero() {
 		if utils.ContainsString(instance.ObjectMeta.Finalizers, HostFinalizerName) {
+
+			// Remove the finalizer so we don't try to do this delete action again.
+			// Defer the removal of the finalizer until the end of the function
+			defer r.removeHostFinalizer(instance)
+
 			// A finalizer is still present so we need to try to delete the
 			// host from the system.
 			if host != nil {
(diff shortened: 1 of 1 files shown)


controllers/host/host_controller.go
```diff
@@ -1877,14 +1877,17 @@ func (r *HostReconciler) ReconcileResource(
 		if utils.ContainsString(instance.ObjectMeta.Finalizers, HostFinalizerName) {
 			// A finalizer is still present so we need to try to delete the
 			// host from the system.
-			if host != nil {
+			if host == nil {
+				logHost.Info("host being deleted is no longer present on system")
+			} else {
 				err = r.ReconcileDeletedHost(client, instance, host)
 				if err != nil {
+					instance.ObjectMeta.Finalizers = utils.RemoveString(instance.ObjectMeta.Finalizers, HostFinalizerName)
```
yjian118, line 1885: Those tasks are identical to line 1894-1897, why to execute them here in error case why they will be eventually triggered later?

vineela-pachchipulusu: If ReconcileDeletedHost method is returning error then the finalizer is not removed through the code
Which is resulting in the stuck of the kubectl delete host 
So before returning if the finalizer is removed the host gets deleted

yjian118: In this case, don't we need to remove the finalizer before deleting the host? Why throw an error out?

vineela-pachchipulusu: we need to delete the finalizer  to execute the kubectl delete host command successfully
SO the code is updated accordingly

controllers/host/host_controller.go
```diff
@@ -1875,22 +1875,29 @@ func (r *HostReconciler) ReconcileResource(
 
 	if !instance.DeletionTimestamp.IsZero() {
 		if utils.ContainsString(instance.ObjectMeta.Finalizers, HostFinalizerName) {
+
+			// Remove the finalizer so we don't try to do this delete action again.
+			// Defer the removal of the finalizer until the end of the function
+			defer func() {
```
yjian118, line 1881: 1. By adding the defer function, it will always triggered in the scope of ReconcileResource, which is wrong, it should be only executed when !instance.DeletionTimestamp.IsZero(), you'll need to create a new method to wrap the section in the if block.
2. You don't need to change the location of the section as it is expected to be always triggered with defer
3. Prefer to have a named function for the defer function for ease to unit test

yjian118 (changes requested):

controllers/host/host_controller.go
vineela-pachchipulusu: The defer statement executes only when !instance.DeletionTimestamp.IsZero() is true
Also it will be executed in the scope of the if block because https://github.com/Wind-River/cloud-platform-deployment-manager/blob/master/controllers/host/host_controller.go#L1904 is returning inside the block

yjian118 (approved):

controllers/host/host_controller.go
```diff
@@ -1875,23 +1875,30 @@ func (r *HostReconciler) ReconcileResource(
...
+				instance.ObjectMeta.Finalizers = utils.RemoveString(instance.ObjectMeta.Finalizers, HostFinalizerName)
+				if err1 := r.Client.Update(context.Background(), instance); err1 != nil {
+					logHost.Error(err, "failed to remove the finalizer in the host because of the error:%v")
```
yjian118, line 1884: err1

vineela-pachchipulusu: Changed it

hugonicodemos (approved):

PR #401, opened 2024-09-06
Update factory install config workflow

yjian118:
Update factory install config workflow
As the factory install workflow changes: starting from fully deployed
system to only the active controller was deployed. This commit updates
the configuration workflow to align with the starting point: using the
day-1 operation rather than day-2 operation to config the system.

This commit:
1. Removes the strategy-related code for configuration post-factory
install.
2. Update all the resources' reconciled status to "false" once to
unblock the reconciliation.
3. Set the configuration post factory install finalized post the
unlock of the active controller.

With this change, the configuration post-factory install should no
longer trigger a vim strategy to orchestrate the lock/unlock of the
hosts.

Additional requests:
1 the host should be locked before applying the configMap to start the
configuration.
2 the administrative status should be set to "unlock" to trigger the
unlock of the host.

Test plan:
1. Deploy an active controller of a DX system, and lock the controller.
2. Apply the configMap to trigger the configuration.
3. Verify the host can be reconfigured and unlocked by DM.
4. Verify the host can be reconciled post-unlock.

### DIFF
--- a/README.md
+++ b/README.md
@@ -529,6 +529,10 @@ expected to be applied once again to finalize the configuration of the system.
 The Deployment Manager can read/update a configmap and orchestrate the
 configuration process accordingly.
 
+Note: only one host is expected to deployed before updating again, and the host
+is expected to be locked. Otherwise, the configMap will be set as finalized
+to block the further operation.
+
 ```yaml
 apiVersion: v1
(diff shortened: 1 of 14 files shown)


hugonicodemos (approved):
lgtm

PR #403, opened 2024-09-20
Enable playbook task for ignoring unreachable servers

sselvara1:
DM playbook fails when the task status is set to unreachable instead of failed. Seems the clause 'ignore_errors: true' is treated different than just a task failure. From ansible 2.7, it is possible to use ignore_unreachable.

Test Cases:
1. PASS: Run dcmanager add from system controller. Deployment config is applied successfully on controller-0. 'deploy status' shall be 'complete' for the subcloud.
2. PASS: Deploy AIO-SX successfully.

### DIFF
--- a/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
+++ b/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
@@ -643,7 +643,7 @@
                 wait_for:
                   timeout: 300
                 register: waiting_after_failures
-                ignore_errors: true
+                ignore_unreachable: true
 
               - name: Get post unlock state
                 shell: >-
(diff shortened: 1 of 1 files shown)


yjian118 (changes requested):
Why do we update the stable branch? Any request to deliver it in a patch? You can share the info directly to me

sselvara1:
> Why do we update the stable branch? Any request to deliver it in a patch? You can share the info directly to me

The Fix is tested on HW lab for 22.12. Hence created this PR on stable branch.
Here is the PR for master branch - 
https://github.com/Wind-River/cloud-platform-deployment-manager/pull/404

sselvara1:
Fix is provided in master branch with these 2 PRs - 

- https://github.com/Wind-River/cloud-platform-deployment-manager/pull/404
- https://github.com/Wind-River/cloud-platform-deployment-manager/pull/399

PR #405, opened 2024-10-02
DM does not reconcile due to the runtime installed certs

vineela-pachchipulusu:
DM should ignore the ssl_ca certificates that gets installed in the system at the runtime

Test Plan:
PASS: The AIO-SX VM should not show delta with certs and host insync should be true when day2 operation is attempted on the host

### DIFF
--- a/controllers/system/system_controller.go
+++ b/controllers/system/system_controller.go
@@ -1080,6 +1080,23 @@ func (r *SystemReconciler) ReconcileSystemFinal(client *gophercloud.ServiceClien
 	return nil
 }
 
+// Fixes extra certs from the response of the current configuration according to the certs specified in the profile.
+func FixCertsToManage(specCerts, currentCerts *starlingxv1.CertificateList) starlingxv1.CertificateList {
+	var res starlingxv1.CertificateList
+	certificateMap := make(map[string]bool)
+
+	//signature of spec certs are empty at this point so secrets are checked rather than signatures
+	for _, cert := range *specCerts {
+		certificateMap[cert.Secret] = true
+	}
+	for _, cert := range *currentCerts {
+		if _, ok := certificateMap[cert.Secret]; ok {
+			res = append(res, cert)
+		}
+	}
+	return res
+}
+
 // ReconcileRequired determines whether reconciliation is allowed/required on
 // the System resource.  Reconciliation is required if there is a difference
 // between the configured Spec and the current system state.  Reconciliation
@@ -1100,6 +1117,12 @@ func (r *SystemReconciler) ReconcileRequired(instance *starlingxv1.System, spec 
 	current, err := starlingxv1.NewSystemSpec(*info)
 	if err != nil {
 		return err, false
+	}
+
+	// We need to remove the runtime installed certificated from the current
+	if spec.Certificates != nil && current.Certificates != nil {
+		res := FixCertsToManage(spec.Certificates, current.Certificates)
+		current.Certificates = &res
 	}
 
 	if spec.DeepEqual(current) {
--- a/controllers/system/system_controller_test.go
+++ b/controllers/system/system_controller_test.go
@@ -163,4 +163,57 @@ var _ = Describe("System controller", func() {
 			Expect(k8sClient.Create(ctx, created)).To(Succeed())
 		})
 	})
+	Context("For the FixCertsToManage func when current has runtimeCerts", func() {
+		It("Should remove the rutime certs and return the remaining", func() {
+			specCerts := &starlingxv1.CertificateList{
+				{
+					Type:      "ssl_ca",
+					Secret:    "ssl-ca-secret-1",
+					Signature: "",
+				},
+				{
+					Type:      "ssl_ca",
+					Secret:    "ssl-ca-secret-2",
+					Signature: "",
+				},
+				{
+					Type:      "ssl_ca",
+					Secret:    "ssl-ca-secret-3",
+					Signature: "",
+				},
+			}
+			currentCerts := &starlingxv1.CertificateList{
+				{
+					Type:      "ssl_ca",
+					Secret:    "ssl-ca-secret-1",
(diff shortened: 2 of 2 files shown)


controllers/system/system_controller.go
```diff
@@ -1374,6 +1374,17 @@ func (r *SystemReconciler) ReconcileResource(client *gophercloud.ServiceClient,
...
+	certList := make([]certificates.Certificate, 0)
+	for _, cert := range systemInfo.Certificates {
+		for _, instCert := range *instance.Spec.Certificates {
+			if cert.Signature == instCert.Signature {
+				certList = append(certList, cert)
+				break
+			}
+		}
+	}
+	systemInfo.Certificates = certList
```
yjian118, line 1386: the update of the systemInfo should stay in systemInfo.PopulateSystemInfo

vineela-pachchipulusu: The parameters of PopulateSystemInfo doesnt have the instance in it
WHich is restricting me to update the PopulateSystemInfo

controllers/system/system_controller.go
```diff
@@ -1374,6 +1374,17 @@ func (r *SystemReconciler) ReconcileResource(client *gophercloud.ServiceClient,
 		return err
 	}
 
+	certList := make([]certificates.Certificate, 0)
+	for _, cert := range systemInfo.Certificates {
+		for _, instCert := range *instance.Spec.Certificates {
```
yjian118, line 1379: The info from the spec is only expected to be used to generate the profile

yjian118 (changes requested):
I believe this commit is a wrong approach, we need the DM to install the certificate if not found. Don't disable this function as it is still required

controllers/system/system_controller.go
```diff
@@ -1102,6 +1102,19 @@ func (r *SystemReconciler) ReconcileRequired(instance *starlingxv1.System, spec
 		return err, false
 	}
 
+	var res starlingxv1.CertificateList
+	certificateMap := make(map[string]bool)
+
+	for _, eachCertificate := range *spec.Certificates {
```
yjian118, line 1108: eachCertificate  can be renamed as "certificate" or "cert"

controllers/system/system_controller.go
```diff
@@ -1102,6 +1102,19 @@ func (r *SystemReconciler) ReconcileRequired(instance *starlingxv1.System, spec
 		return err, false
 	}
 
+	var res starlingxv1.CertificateList
```
yjian118, line 1105: Expect a new method to wrap the new codes, and a unit test against it

yjian118: Need to add description here to explain why not add the other certs in the map to the current list

controllers/system/system_controller.go
```diff
@@ -1102,6 +1102,19 @@ func (r *SystemReconciler) ReconcileRequired(instance *starlingxv1.System, spec
...
+	certificateMap := make(map[string]bool)
+
+	for _, eachCertificate := range *spec.Certificates {
+		certificateMap[eachCertificate.Secret] = true
```
yjian118, line 1109: Any reason why comparing based on the secret rather than the signature?

controllers/system/system_controller.go
vineela-pachchipulusu: updated

controllers/system/system_controller.go
vineela-pachchipulusu: I have updated accordingly

controllers/system/system_controller.go
vineela-pachchipulusu: Because by that point spec doesn't get the filled with the corresponding signatures
Signatures will be empty values in the spec

controllers/system/system_controller.go
```diff
@@ -1080,6 +1080,22 @@ func (r *SystemReconciler) ReconcileSystemFinal(client *gophercloud.ServiceClien
 	return nil
 }
 
+// Removes the extra certs from the currentCerts to the specCerts
+func RemoveRuntimeCerts(specCerts, currentCerts *starlingxv1.CertificateList) starlingxv1.CertificateList {
+	var res starlingxv1.CertificateList
+	certificateMap := make(map[string]bool)
+
+	for _, cert := range *specCerts {
+		certificateMap[cert.Secret] = true
```
yjian118, line 1089: Can you add a comment regarding why to check based on the secret rather than the signature?

controllers/system/system_controller.go
yjian118: Suggest to add a comment in the new method to explain that.

yjian118:
Expect test plan and the execution result

controllers/system/system_controller.go
```diff
@@ -1080,6 +1080,22 @@ func (r *SystemReconciler) ReconcileSystemFinal(client *gophercloud.ServiceClien
 	return nil
 }
 
+// Removes the extra certs from the currentCerts to the specCerts
+func RemoveRuntimeCerts(specCerts, currentCerts *starlingxv1.CertificateList) starlingxv1.CertificateList {
```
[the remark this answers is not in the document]
vineela-pachchipulusu: Updated

controllers/system/system_controller.go
```diff
@@ -1080,6 +1080,22 @@ func (r *SystemReconciler) ReconcileSystemFinal(client *gophercloud.ServiceClien
 	return nil
 }
 
+// Removes the extra certs from the currentCerts to the specCerts
```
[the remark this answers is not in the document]
vineela-pachchipulusu: Updated

controllers/system/system_controller.go
vineela-pachchipulusu: Explained

hugonicodemos (approved):

yjian118 (approved):

PR #409, opened 2024-10-29
Fix for DM failure: Std subcloud installation

MahalakshmiManthiram:
Playbook fails when it tries to reboot and subsequent reboot was not executed.
Required conditions were added to execute the second reboot task.

Test plan:
Deployment of subcloud3 was failing at the reboot task. With this fix,
verified the reboot task is getting failed and subsequent reboot is executed
successfully.

### DIFF
--- a/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
+++ b/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
@@ -743,7 +743,19 @@
             wait_for:
               timeout: "{{ boot_wait_time }}"
             register: waiting_after_reboot
+            # Unreachable: Ignore a task failure due to the host instance being ’UNREACHABLE’ 
+            # with the ignore_unreachable keyword. Ansible ignores the task errors but continues
+            # to execute future tasks against the unreachable host. This usually happens 
+            # if the host is network-inaccessible. For example, during a second reboot that
+            # temporarily disrupts SSH connectivity, Ansible will mark it as unreachable since 
+            # it can’t connect to the host at all.
             ignore_unreachable: true
+            # Failed: It can ignore errors when the task is failed, except those syntax errors, 
+            # undefined variable errors, connection failures, execution issues.
+            # This occurs if Ansible can reach the host, but the task itself encounters an issue. This 
+            # could happen if the host is reachable but not fully ready to respond, or if a required
+            # port isn’t open yet.
+            ignore_errors: true
 
           # Retry block: sometimes system reboots twice
           # It will take some extra time.
@@ -760,7 +772,7 @@
               delay: 20
 
             # Waiting task after unlock to catch the right status
-            - name: Wait for {{ boot_wait_time }} seconds to ensure not affecting host status
+            - name: Waiting after second reboot for {{ boot_wait_time }} seconds to ensure not affecting host status
               wait_for:
                 timeout: "{{ boot_wait_time }}"
               register: waiting_after_new_reboot
@@ -774,7 +786,7 @@
               register: new_reboot
               when: waiting_after_new_reboot.failed
 
-            when: waiting_after_reboot.failed
+            when: waiting_after_reboot.failed or waiting_after_reboot.unreachable
 
           - name: Set retries to check resource reconciled status for simplex
             set_fact:


docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -774,7 +775,7 @@
               register: new_reboot
               when: waiting_after_new_reboot.failed
 
-            when: waiting_after_reboot.failed
+            when: waiting_after_reboot.failed or waiting_after_new_reboot.unreachable
```
yjian118, line 778: waiting_after_new_reboot is defined in line767, the condition should not contain a register inside the block.
Please mark as draft before tested

yjian118:
I don't believe the test hits the change path.

yjian118 (changes requested):

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -744,6 +744,7 @@
               timeout: "{{ boot_wait_time }}"
             register: waiting_after_reboot
             ignore_unreachable: true
+            ignore_errors: true
```
yjian118, line 747: Please add comment here to explain why ignore_unreachable and ignore_errors are both expected

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
yjian118, line 778: Further, I believe the previous change regarding the "ignore_unreachable" of waiting_after_reboot was not properly tested.

yjian118, line 778: Suggest to change the task name of line 764 to avoid two tasks with a same name - confuse to read the log

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
MahalakshmiManthiram: Different scenarios were seen and observed that the task could be marked as failed sometimes and could be marked as unreachable. It could be something specific of ansible that we might need to investigate further. But in this case, with both conditions, we cover the  possible scenarios.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
MahalakshmiManthiram: Changes are done

MahalakshmiManthiram: It was tested after changing the condition to when: waiting_after_reboot.failed or waiting_after_reboot.unreachable

MahalakshmiManthiram: Condition was corrected to when: waiting_after_reboot.failed or waiting_after_reboot.unreachable and tested successfully.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
yjian118: Can you add two comments regarding the cases "ignore_unreachable" and "ignore_errors" respectively for future reference and understanding?

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -774,7 +775,7 @@
               register: new_reboot
               when: waiting_after_new_reboot.failed
```
yjian118, line 776: Do we need to consider ignore_unreachable in this case?

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
MahalakshmiManthiram: It can ignore a task failure due to the host instance being ’UNREACHABLE’ with the ignore_unreachable keyword. Ansible ignores the task errors but continues to execute future tasks against the unreachable host.
It can ignore errors when the task is failed, except those syntax errors, undefined variable errors, connection failures, execution issues.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
MahalakshmiManthiram: i think not required for now.

MahalakshmiManthiram:
About the cases "ignore_unreachable" and "ignore_errors" for future reference-

Unreachable: 
It can ignore a task failure due to the host instance being ’UNREACHABLE’ with the ignore_unreachable keyword. Ansible ignores the task errors but continues to execute future tasks against the unreachable host.
This usually happens if the host is network-inaccessible. For example, during a second reboot that temporarily disrupts SSH connectivity, Ansible will mark it as unreachable since it can’t connect to the host at all.

Failed: It can ignore errors when the task is failed, except those syntax errors, undefined variable errors, connection failures, execution issues.
This occurs if Ansible can reach the host, but the task itself encounters an issue. This could happen if the host is reachable but not fully ready to respond, or if a required port isn’t open yet.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -744,6 +744,7 @@
               timeout: "{{ boot_wait_time }}"
             register: waiting_after_reboot
             ignore_unreachable: true
```
MahalakshmiManthiram, line 746: About the cases "ignore_unreachable" and "ignore_errors" for future reference-

Unreachable:
It can ignore a task failure due to the host instance being ’UNREACHABLE’ with the ignore_unreachable keyword. Ansible ignores the task errors but continues to execute future tasks against the unreachable host.
This usually happens if the host is network-inaccessible. For example, during a second reboot that temporarily disrupts SSH connectivity, Ansible will mark it as unreachable since it can’t connect to the host at all.

Failed: It can ignore errors when the task is failed, except those syntax errors, undefined variable errors, connection failures, execution issues.
This occurs if Ansible can reach the host, but the task itself encounters an issue. This could happen if the host is reachable but not fully ready to respond, or if a required port isn’t open yet.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
MahalakshmiManthiram, line 747: About the cases "ignore_unreachable" and "ignore_errors" for future reference-

Unreachable:
It can ignore a task failure due to the host instance being ’UNREACHABLE’ with the ignore_unreachable keyword. Ansible ignores the task errors but continues to execute future tasks against the unreachable host.
This usually happens if the host is network-inaccessible. For example, during a second reboot that temporarily disrupts SSH connectivity, Ansible will mark it as unreachable since it can’t connect to the host at all.

Failed: It can ignore errors when the task is failed, except those syntax errors, undefined variable errors, connection failures, execution issues.
This occurs if Ansible can reach the host, but the task itself encounters an issue. This could happen if the host is reachable but not fully ready to respond, or if a required port isn’t open yet.

MahalakshmiManthiram, line 747: Errors: It can ignore errors when the task is failed, except those syntax errors, undefined variable errors, connection failures, execution issues.
This occurs if Ansible can reach the host, but the task itself encounters an issue. This could happen if the host is reachable but not fully ready to respond, or if a required port isn’t open yet.

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -743,7 +743,21 @@
             wait_for:
               timeout: "{{ boot_wait_time }}"
             register: waiting_after_reboot
+
```
yjian118, line 746: Can you remove the blank line 746

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -743,7 +743,21 @@
...
+            # Unreachable: Ignore a task failure due to the host instance being ’UNREACHABLE’ 
+            # with the ignore_unreachable keyword. Ansible ignores the task errors but continues
+            # to execute future tasks against the unreachable host. This usually happens 
+            # if the host is network-inaccessible. For example, during a second reboot that
+            # temporarily disrupts SSH connectivity, Ansible will mark it as unreachable since 
+            # it can’t connect to the host at all.
             ignore_unreachable: true
+
```
yjian118, line 754: Can you remove the blank line 754

yjian118 (changes requested):
Expect a re-test with the latest commit

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
MahalakshmiManthiram: Done

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
MahalakshmiManthiram: Done

yjian118 (approved):

PR #410, opened 2024-10-30
Remove status.delta when controller is insync

wasnio:
DM must not have status.delta when controllers are insync. This commit
ensures no delta always controllers are in sync after a day-2 update.

Test Case:
PASS - perform day-2 updates, changing parameters in the deployment
configuration and ensure no delta after insync true

### DIFF
--- a/controllers/host/host_controller.go
+++ b/controllers/host/host_controller.go
@@ -1774,36 +1774,18 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
 	}
 
 	logHost.Info("comparing profile attributes", "host", host.ID)
-	inSync := r.CompareAttributes(profile, current, instance, host.Personality, system_info)
+	instance.Status.InSync = r.CompareAttributes(profile, current, instance, host.Personality, system_info)
+	common.SetInstanceDelta(instance, profile, current, common.HostProperties, r.Client.Status(), logHost)
 
-	// Continue the following steps even inSync:
 	// strategy not finished
-	if inSync && (instance.Status.StrategyRequired == cloudManager.StrategyNotRequired) {
+	if instance.Status.InSync && (instance.Status.StrategyRequired == cloudManager.StrategyNotRequired) {
 		logHost.V(2).Info("no changes between composite profile and current config", "host", host.ID)
-		instance.Status.Delta = ""
 		return nil
 	}
 
 	logHost.Info("defaults are:", "values", defaults)
-
 	logHost.Info("final profile is:", "values", profile)
-
 	logHost.Info("current config is:", "values", current)
-
-	deltaString, err := common.GetDeltaString(profile, current, common.HostProperties)
-	if err != nil {
-		logHost.Info(fmt.Sprintf("failed to get Delta status:  %s\n", err))
-	}
-
-	if deltaString != "" {
-		logHost.V(2).Info(fmt.Sprintf("delta configuration:%s\n", deltaString))
-		instance.Status.Delta = deltaString
-
-		err = r.Client.Status().Update(context.TODO(), instance)
-		if err != nil {
-			logHost.Info(fmt.Sprintf("failed to update status: %s", err))
-		}
-	}
 
 	if instance.Status.Reconciled &&
 		r.StopAfterInSync() &&
--- a/controllers/system/system_controller.go
+++ b/controllers/system/system_controller.go
@@ -1125,29 +1125,12 @@ func (r *SystemReconciler) ReconcileRequired(instance *starlingxv1.System, spec 
 		current.Certificates = &res
 	}
 
-	if spec.DeepEqual(current) {
-		logSystem.V(2).Info("no changes between spec and current configuration")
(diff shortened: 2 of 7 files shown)


controllers/host/host_controller.go
```diff
@@ -1730,12 +1730,32 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
 
 	inSync := r.CompareAttributes(profile, current, instance, host.Personality)
 
+	if inSync {
+		logHost.Info("Cleaning Delta since inSync")
+		instance.Status.Delta = ""
```
yjian118, line 1735: Updating the intance may not sufficient, need to consider update the status as line 1744

wasnio: done

controllers/host/host_controller.go
```diff
@@ -1729,37 +1729,36 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
 	}
 
 	inSync := r.CompareAttributes(profile, current, instance, host.Personality)
+	oldDelta := instance.Status.Delta
+
+	if inSync {
```
yjian118, line 1734: The logic here looks good, can we wrap it into a new method? I'm wondering if we need to care about the other controllers. And I'll prefer a new unit test against the new method if possible

controllers/system/system_controller.go
```diff
@@ -1125,30 +1125,33 @@ func (r *SystemReconciler) ReconcileRequired(instance *starlingxv1.System, spec
...
-		instance.Status.Delta = deltaString
+	if oldDelta != instance.Status.Delta {
+		logSystem.Info(fmt.Sprintf("delta configuration:%s\n", instance.Status.Delta))
 		err = r.Client.Status().Update(context.TODO(), instance)
 		if err != nil {
 			logSystem.Info(fmt.Sprintf("failed to update status:  %s\n", err))
 		}
 	}
 
+	if inSync {
```
yjian118, line 1147: Not prefer to return because of insync, lin1155 should be enough to handle this case

controllers/system/system_controller.go
```diff
@@ -1125,30 +1125,33 @@ func (r *SystemReconciler) ReconcileRequired(instance *starlingxv1.System, spec
...
+		logSystem.V(2).Info("no changes between spec and current configuration")
+		return nil, false
+	}
+
+	logSystem.Info("spec is:", "values", spec)
```
yjian118, line 1152: This log was already elsewhere, can you make it v2 or remove it?

yjian118 (changes requested):
@enzocandotti can you help to review this commit as you are the main introducer of the delta status

controllers/system/system_controller.go
wasnio: done

controllers/host/host_controller.go
wasnio: made it generic. I'll raise a new issue to replace the other controllers status.delta with the SetInstanceDelta function

enzocandotti:
Please, add an explanation about how the workflow would change by adding this commit.

enzocandotti:
@wasnio have you seen any other example of a Delta not being cleared when the resource is in-sync, aside from the administrativeState? It seems like a major change to the workflow that we should consider carefully to see if it’s truly necessary.

If deltaString != "", it's because the configurations are that way.
If there’s no other case where a Delta is not empty while the resource is in-sync, aside from the administrativeState, maybe we could ignore the administrativeState when checking for the Delta.

enzocandotti:
@yjian118 I’d like to get your input on this. I understand that the Delta status helps users identify which configurations need to be applied and explains why the resource isn’t in-sync. It makes sense for the Delta to be empty when the resource is in-sync. But in this case, where there are configuration differences (non-empty Delta) and the resource is in-sync, does this mean the resource was able to synchronize even with those differences?

Do we actually have an issue by showing the configuration differences?

We should consider whether we need to continue with this workflow of forcibly clearing the Delta when the resource is in-sync, or perhaps ignore certain differences that, while still present, allow the resource to be in-sync. @wasnio could you please show us the Host and Datanetworks delta strings that you saw?

In my opinion, this is a significant change at this time for a minor issue, and we should fully understand and test it with all Day-2 scenarios.

hugonicodemos (approved):

controllers/system/system_controller.go
```diff
@@ -1125,29 +1125,8 @@ func (r *SystemReconciler) ReconcileRequired(instance *starlingxv1.System, spec
...
 
-	if spec.DeepEqual(current) {
-		logSystem.V(2).Info("no changes between spec and current configuration")
-		instance.Status.Delta = ""
-		return nil, false
-	}
-
-	logSystem.Info("spec is:", "values", spec)
-
-	logSystem.Info("current is:", "values", current)
```
enzocandotti, line 1136: Can we keep these logs?

enzocandotti:
Just a comment for future traceability, this change was implemented just for Host and System resources.

PR #413, opened 2024-11-08
Fixing storage.filesystem syncrhonization for AIO-DX

fsanches-wr:
The Storage.Filesystem synchronization is failing for AIO-DX because the 'ceph' host filesystem is created during the first unlock of the controllers. The host profile must not have the 'ceph' filesystem, but the current config will have it after the unlock, because the sysinv will create that automatically to support the 3 Ceph monitors.

This fix will ignore the 'ceph' filesystem synchronization when this is an AIO-DX deployment.

Test-Plan
  PASS: Deploy an AIO-DX and check both controllers are InSync and
        Reconciled
  PASS: Deploy a Standard and check all the hosts are InSync and
        Reconciled

### DIFF
--- a/controllers/host/host_controller.go
+++ b/controllers/host/host_controller.go
@@ -938,13 +938,13 @@ func (r *HostReconciler) CompareOSDs(in *starlingxv1.HostProfileSpec, other *sta
 // CompareAttributes determines if two profiles are identical for the
 // purpose of reconciling a current host configuration to its desired host
 // profile.
-func (r *HostReconciler) CompareAttributes(in *starlingxv1.HostProfileSpec, other *starlingxv1.HostProfileSpec, instance *starlingxv1.Host, personality string) bool {
+func (r *HostReconciler) CompareAttributes(in *starlingxv1.HostProfileSpec, other *starlingxv1.HostProfileSpec, instance *starlingxv1.Host, personality string, system_info *cloudManager.SystemInfo) bool {
 	// This could be replaced with in.DeepEqual(other) but it is coded this way
 	// (and tested this way) to ensure that if both the "enabled" and "disabled"
 	// comparisons are true then no reconciliation is missed.  The intent is
 	// that CompareEnabledAttributes && CompareDisabledAttributes will always
 	// be equivalent to DeepEqual.
-	return r.CompareEnabledAttributes(in, other, instance, personality) &&
+	return r.CompareEnabledAttributes(in, other, instance, personality, system_info) &&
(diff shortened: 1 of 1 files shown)


controllers/host/host_controller.go
```diff
@@ -979,12 +979,41 @@ func (r *HostReconciler) CompareEnabledAttributes(in *starlingxv1.HostProfileSpe
...
 		if in.Storage != nil && in.Storage.FileSystems != nil {
-			if other.Storage == nil {
+			if other.Storage == nil || other.Storage.FileSystems == nil {
 				return false
 			}
 
-			if !in.Storage.FileSystems.DeepEqual(other.Storage.FileSystems) {
-				return false
+			// Special case: 'ceph' filesystem must be ignored for All-in-one Duplex.
+			if system_info.SystemType == cloudManager.SystemTypeAllInOne && system_info.SystemMode != cloudManager.SystemModeSimplex {
```
yjian118, line 987: Question: does it only apply on AIODX or AIODX+?

yjian118 (approved):

controllers/host/host_controller.go
```diff
@@ -979,12 +979,41 @@ func (r *HostReconciler) CompareEnabledAttributes(in *starlingxv1.HostProfileSpe
...
+					// Skip ceph filesystem
+					if fsInfo.Name == "ceph" {
+						continue
+					}
+
+					// Compare 'in' and 'other' fileystems
+					found := false
+					for _, fs := range *in.Storage.FileSystems {
+						if fs.Name != fsInfo.Name {
+							continue
```
hugonicodemos, line 998: Can you extract the inner `for` loop to a method: findFileSystemByName(*in.Storage.FileSystems, fsInfo.Name) and we can do something like:

if matchedFileSystem == nil || matchedFs.Size != fsInfo.Size {
        return false
}

this will eliminate the need for the flag and make the code more modular. We want to avoid expanding the method’s scope, as that would make it more complex to understand.

hugonicodemos (changes requested):
Just a minor suggestion, besides that Lgtm

controllers/host/host_controller.go
fsanches-wr: I can do that.
I did this way copying from another part of the code. My intention was to maintain the same schema.

hugonicodemos: @fsanches-wr This was a suggestion to improve the code. Since it’s not a major change to the method, I’ll go ahead and merge it.

hugonicodemos (approved):

PR #420, opened 2024-11-14
Add logs on host reconcile path

enzocandotti:
This commits aims to add more logs in the host reconcile path since it was seen situations in which the reconciler is blocked and current logs can not give the enough information to debug it.

### DIFF
--- a/controllers/host/host_controller.go
+++ b/controllers/host/host_controller.go
@@ -2334,6 +2334,7 @@ func (r *HostReconciler) Reconcile(ctx context.Context, request ctrl.Request) (r
 	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation &&
 		instance.Status.Reconciled &&
 		instance.Status.DeploymentScope == "bootstrap" &&
+		*instance.Status.AvailabilityStatus == "available" &&
 		instance.Status.StrategyRequired == cloudManager.StrategyNotRequired &&
 		!platformnetwork_update_required {
 
--- a/controllers/host/host_controller.go
+++ b/controllers/host/host_controller.go
@@ -1564,9 +1564,12 @@ func (r *HostReconciler) StopAfterInSync() bool {
 // ReconcileExistingHost is responsible for dealing with the provisioning of an
 // existing host.
 func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient, instance *starlingxv1.Host, profile *starlingxv1.HostProfileSpec, host *hosts.Host, reqNs string) error {
+	logHost.Info("Starting to configure on existing host", "host", host.ID)
+
 	var defaults *starlingxv1.HostProfileSpec
 	var current *starlingxv1.HostProfileSpec
 	var platform_network_subreconciler_errs []error
+
 	if !host.Stable() {
 		msg := "waiting for a stable state for existing host"
 		m := NewStableHostMonitor(instance, host.ID)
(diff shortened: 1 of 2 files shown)


controllers/host/host_controller.go
```diff
@@ -1564,106 +1564,124 @@ func (r *HostReconciler) StopAfterInSync() bool {
...
 			m := NewInventoryCollectedMonitor(instance, host.ID)
 			return r.CloudManager.StartMonitor(m, msg)
 		}
 
-		logHost.Info("collecting default values")
-
+		logHost.Info("collecting default values for host", "host", host.ID)
 		defaults, err = r.BuildHostDefaults(instance, hostInfo)
 		if err != nil {
+			logHost.Error(err, "failed to build host defaults", "host", host.ID)
```
yjian118, line 1655: I am a bit concerned about the pattern in this commit like:
log Error
return err

If an err is returned to the previous level, it should be logged there, unless you want to change/wrap the error message, otherwise it will be logged twice, which is not a good practice

controllers/host/host_controller.go
```diff
@@ -1564,106 +1564,124 @@ func (r *HostReconciler) StopAfterInSync() bool {
...
 func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient, instance *starlingxv1.Host, profile *starlingxv1.HostProfileSpec, host *hosts.Host, reqNs string) error {
+	logHost.Info("starting reconciliation for existing host", "host", host.ID)
+
 	var defaults *starlingxv1.HostProfileSpec
 	var current *starlingxv1.HostProfileSpec
 	var platform_network_subreconciler_errs []error
+
 	if !host.Stable() {
 		msg := "waiting for a stable state for existing host"
+		logHost.Info("host is not stable", "host", host.ID)
```
yjian118, line 1575: The msg will be recorded in the StartMonitor, why do we need another similar message?

controllers/host/host_controller.go
```diff
@@ -1564,106 +1564,124 @@ func (r *HostReconciler) StopAfterInSync() bool {
 // ReconcileExistingHost is responsible for dealing with the provisioning of an
 // existing host.
 func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient, instance *starlingxv1.Host, profile *starlingxv1.HostProfileSpec, host *hosts.Host, reqNs string) error {
+	logHost.Info("starting reconciliation for existing host", "host", host.ID)
```
yjian118, line 1567: Starting to configure on existing host

yjian118 (changes requested):

controllers/host/host_controller.go
enzocandotti: Removed all logs that are printed in StartMonitor

controllers/host/host_controller.go
enzocandotti: Done

yjian118 (changes requested):
a tiny update is expected

controllers/host/host_controller.go
```diff
@@ -1703,7 +1711,7 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
 
 	// N3000 interface name change apply
 	if host.IsUnlockedEnabled() {
-		logHost.Info("Sync interface name")
+		logHost.Info("syncing interface name for N3000", "host", host.ID)
```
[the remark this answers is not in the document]
enzocandotti: Done

yjian118:
1. There's a merge conflict to resolve
2. We'll need to test the build of the app-dm repo after introducing @wasnio commit which introduced the "slice" lib may not work in go 1.19

yjian118 (approved):

PR #427, opened 2024-11-27
Delete existing system resource during enrollment

yjian118:
During the enrollment, it is possible that the system resource name applied during enrollment deffers from the existing one. To prevent this problem, this commit checks the existing system resources and delete them before applying new system resource.

Test plan:
Passed -  successfully finish the subcloud enrollment with this change. The system resource applied during the enrollment deffers from the existing system resource. Verify the system resource can be insync and reconciled after enrollment.


wasnio (approved):

hugonicodemos (approved):
LGTM

enzocandotti (approved):

PR #428, opened 2024-11-27
Remove test target for docker-build in Makefile

wasnio:
This commit removes the test target in the docker-build to avoid running tests twice when the user issues make command.

Test Case:
- make, tests must run before build
- make docker-build, tests must not run

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,7 @@ else
 endif
 
 .PHONY: docker-build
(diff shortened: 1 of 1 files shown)


hugonicodemos (approved):

yjian118 (approved):

yjian118:
Can we update it to the master as well?

PR #436, opened 2024-12-16
Deployment manager failed due to BMC secret

MahalakshmiManthiram:
Serializing the IncompleteSecret to generate correct configuration. Updated the test cases.

Test plan:
- 'make' and 'make test' ran successfully
- 'deployctl tool generates the expected configuration

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -95,9 +95,12 @@ func parseIncompleteSecret(secret *v1.Secret) *IncompleteSecret {
 	warningMsg := "Warning: Incomplete secret, please replace it with the secret content"
 	if secret.Type == v1.SecretTypeTLS {
 		return &IncompleteSecret{
-			TypeMeta:   secret.TypeMeta,
-			ObjectMeta: secret.ObjectMeta,
-			Type:       secret.Type,
+			TypeMeta: metav1.TypeMeta{},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      secret.Name,
+				Namespace: secret.Namespace,
+			},
+			Type: secret.Type,
 			Data: map[string]string{
 				v1.TLSCertKey:              warningMsg,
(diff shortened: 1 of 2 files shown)


build/build.go
```diff
@@ -95,9 +95,12 @@ func parseIncompleteSecret(secret *v1.Secret) *IncompleteSecret {
 	warningMsg := "Warning: Incomplete secret, please replace it with the secret content"
 	if secret.Type == v1.SecretTypeTLS {
```
wasnio, line 96: can you add to the yaml encode test case the SecretTypeBasicAuth and IncompleteSecret? Actually only the generic case is covered.

build/build_suite_test.go
```diff
@@ -297,10 +296,7 @@ var _ = Describe("Test Build utilities:", func() {
 				}
 				got := parseIncompleteSecret(&secret)
 				expect := IncompleteSecret{
-					TypeMeta: metav1.TypeMeta{
```
wasnio, line 300: I don't think we need to change here

build/build.go
```diff
@@ -95,9 +95,12 @@ func parseIncompleteSecret(secret *v1.Secret) *IncompleteSecret {
...
 		return &IncompleteSecret{
-			TypeMeta:   secret.TypeMeta,
-			ObjectMeta: secret.ObjectMeta,
-			Type:       secret.Type,
+			TypeMeta: metav1.TypeMeta{},
```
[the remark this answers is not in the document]
MahalakshmiManthiram: This line was removed and tested.

build/build_suite_test.go
MahalakshmiManthiram: Removed the changes and tested.

wasnio (approved):

PR #437, opened 2024-12-18
Deployctl does not generate clusterhost platform network

vineela-pachchipulusu:
This PR implements the generation of cluster-host type platformNetworks via deployctl without --minimal-config


hugonicodemos:
Hi @vineela-pachchipulusu , can you test a deployment with the generated config using cluster-host?

hugonicodemos (approved):

PR #439, opened 2025-01-24
Revert certificate CRD to allow historical cert types

yjian118:
During the upgrade from a previous release to the current release to the 24.09 release, the migration process is using B&R the custom resource rather than using helm to controll them. The legacy cert type needs to be preserved during the apply in order to accept the historical data with additional cert type.

This commit add the legacy cert types back in helm chart to unblock the addmission, however the legacy certs are still expected to be deleted by clean_deprecated_certificates.

Test:
1. Build an image and fresh deploy 24.09.1 system
2. Upgrade from 22.12 to 24.09 w/o failure


yjian118:
We may not need this commit as: addressing the cert in migration script

enzocandotti (approved):

hugonicodemos (approved):

PR #443, opened 2025-02-13
Fix ready state notification race condition in system controller

wasnio:
When system controller is ready, it will notify all the other reconcilers
about the ready state, so the other controllers can start the
reconciliation loop. This commit ensures that the system controller
can notify the other controller even if the last iteration failed to
call the webhook, but the system is reconciled.

Test Case:
- PASS: Deploy AIO-SX

### DIFF
--- a/controllers/system/system_controller.go
+++ b/controllers/system/system_controller.go
@@ -1705,8 +1705,7 @@ func (r *SystemReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 		return ctrl.Result{}, nil
 	}
 
-	err, scope_updated := r.UpdateDeploymentScope(instance)
-	if err != nil {
+	if err, _ := r.UpdateDeploymentScope(instance); err != nil {
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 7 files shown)


controllers/system/system_controller.go
```diff
@@ -1724,12 +1724,26 @@ func (r *SystemReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 	// The status reaches its desired status post reconciled
 	if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation &&
```
yjian118, line 1725: I truly hate the implementation of this block in: https://github.com/Wind-River/cloud-platform-deployment-manager/pull/352, unfortunately, I'm not confident to remove this whole block for now as DM migration concerns. Can you add a TODO above "if instance.Status.ObservedGeneration == instance.ObjectMeta.Generation &&" on all controllers  to indicate we'll remove this block once no longer supporting the migration from helm application to FluxCD application? I'm Okay with this commit for now

yjian118 (approved):

yjian118:
@MahalakshmiManthiram  I'm curious why do we need this block to avoid the reboot of a host during restoration? Only the host controllers are involved in triggering an unlock right? Can we remove the restore annotation and the stop reconcile block in all the controllers other than the host controller?

MahalakshmiManthiram:
> @MahalakshmiManthiram I'm curious why do we need this block to avoid the reboot of a host during restoration? Only the host controllers are involved in triggering an unlock right? Can we remove the restore annotation and the stop reconcile block in all the controllers other than the host controller?

@yjian118 The code block was added across all the controllers for restoring DM Objects. It ensures that DM objects are properly restored without conflicts. It halts the reconciliation process to safely restore the necessary objects and avoid any unintended operations that might affect the restoration.

yjian118 (approved):

hugonicodemos (approved):

PR #445, opened 2025-03-11
Enable the creation of two data VLAN with same VID

wasnio:
This commit fixes the vlan interface search, searching the interface by name and VID to allow the creation of two VLAN with same VID when using two different physical interfaces.

Test Case:
- PASS: configure and verify two VLAN using the same VID


yjian118 (approved):

enzocandotti (approved):

enzocandotti:
Just a minor comment not related to your review, the Copyright date of info.go is not updated. Could you do it?

PR #450, opened 2025-04-09
Improve multi node subcloud availability check

wasnio:
This commit improves the multinode subcloud availability check by checking the inventory for hosts with administrative state equal to unlock and availability other than online and available.

Test Plan:
- PASS: deploy a multinode subcloud


docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -833,18 +833,23 @@
...
+            - debug:
+                msg: >
+                  Retry for {{ host_offline_retries }} seconds until all hosts
+                  can be configured.
+
+            - name: Check hosts cannot be configured by the DM
+              shell: |
+                source /etc/platform/openrc;
+                system host-list --format value |
+                awk '$4 != "locked" && ($6 != "available" && $6 != "online") {print $2}'
```
yjian118, line 852: Sorry, I think the logic here is wrong: we may need to check if the $4=="locked" && ($6 != "available" && $6 != "online")

yjian118 (changes requested):
I feel there's a logic error in my previous draft, would prefer to correct that, others lgtm

yjian118 (approved):

enzocandotti (approved):

PR #451, opened 2025-04-29
Do not raise an error when system instance.reconciled  is true

wasnio:
When instance.Reconciled is true, system controller was raising an error [1]. This commit changes ReconcileRequired to return `false` when the system is reconciled and StopAfterinsync() is `true`.

[1]: https://github.com/Wind-River/cloud-platform-deployment-manager/blob/372578da4d35eb5de0240471078731237cbc2471/controllers/system/system_controller.go#L1135

Test Cases:
- PASS: deploy an AIO-SX
- PASS: `kubectl -n deployment get system` and verify InSync is `true`


hugonicodemos (approved):
LGTM

PR #452, opened 2025-05-01
Delay labels error return until after cpu config

eliud-kyale:
The LabelsReconcile could fail because of cpu
configuration. Since, Labels are reconciled first
we shall delay the error response until
Processors/CPUs have been reconciled
So that the next round of reconciliation will pass labels reconcile

Test plan:

PASS - Deploy AIO-SX
        with stalld host labels with application
        isolated cpus .

### DIFF
--- a/controllers/system/system_controller.go
+++ b/controllers/system/system_controller.go
@@ -1103,7 +1103,7 @@ func FixCertsToManage(specCerts, currentCerts *starlingxv1.CertificateList) star
 // is only allowed if the resource has not already been successfully reconciled
 // at least once; or the user has overridden this check by adding an annotation
 // on the resource.
-func (r *SystemReconciler) ReconcileRequired(instance *starlingxv1.System, spec *starlingxv1.SystemSpec, info *v1info.SystemInfo) (err error, required bool) {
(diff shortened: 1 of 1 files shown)


controllers/host/host_controller.go
```diff
@@ -805,9 +805,10 @@ func (r *HostReconciler) ReconcileDisabledHost(client *gophercloud.ServiceClient
 		return err
 	}
 
-	err = r.ReconcileLabels(client, instance, profile, host)
```
wasnio, line 808: I think it would be better to move this portion to line 837. I think this will fix the problem and can avoid a new reconciliation loop since ReconcileLabels will work after ReconcileProcessors

wasnio (changes requested):

controllers/host/host_controller.go
eliud-kyale: The concern is whether the order matters , if I move the order will it affect other areas ?

wasnio: I think ptpinstnaces relies on labels and you can move with reconcilelables down. I'm not sure about the other one, but I don't think reconcile kernel relies on labels.

PR #468, opened 2025-06-27
Improve get deployment namespace task

wasnio:
This commit updates the "get deployment namespace" task to a more robust approach. The new approach handles cases where the user adds an empty item in the yam list (--- at the end of the file) and when the defined resource does not have the "kind" key.

Test Cases:
- PASS: deploy using a custom namespace
- PASS: deploy using the default namespace

### DIFF
--- a/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
+++ b/docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
@@ -27,21 +27,24 @@
         wait_for_dm_unlock: " {{ wait_for_dm_unlock| default(5) }}"
       when: deployment_config is defined
 
-    - block:
-        - name: Set deployment namespace name fact
-          set_fact:
-            deployment_namespace: >-
-              {{ lookup('file', '{{ deploy_config }}')
-              | from_yaml_all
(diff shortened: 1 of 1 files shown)


docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -27,21 +27,24 @@
...
+      shell: |
+        import yaml
+        with open("{{ deploy_config }}") as f:
+            resources = list(yaml.safe_load_all(f))
+        namespace = "deployment"
+        for resource in resources:
+            if isinstance(resource, dict) and resource.get("kind") == "Namespace":
+                namespace = resource["metadata"]["name"]
+                break
+        print(namespace.strip())
```
hugonicodemos, line 40: I believe we can remove this print

hugonicodemos (changes requested):

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
wasnio: We cannot because this is the output of the script

PR #469, opened 2025-07-04
Skip get deployment namespace task if deploy_config is missing

wasnio:
Test Plan:
PASS: execute the playbook omitting deploy_config variable, playbook
      must ignore the task and install dm.
PASS: execute the playbook providing deploy_config variable, playbook
      must install dm and apply the deployment configuration.


yjian118 (changes requested):
This commit lgtm, however, the deploy_config set by deployment_config is really redundant and confusing, can we change all the "deploy_config" to "deployment_config", and remove the task to set that fact?

enzocandotti (approved):

PR #470, opened 2025-07-16
Allow platform memory configuration on controllers

wasnio:
Test Cases:
TBD


controllers/host/host_controller.go
```diff
@@ -819,11 +819,14 @@ func (r *HostReconciler) ReconcileDisabledHost(client *gophercloud.ServiceClient
 			return err
 		}
 
-		err = r.ReconcileMemory(client, instance, profile, host)
-		if err != nil {
-			return err
-		}
+	}
```
yjian118, line 822: My question to this review is mainly:
As described in line 814: the following configuration is only expected to nodes with worker function(not sure any further development after the initial code submission).

If the memory can be configured, I don't understand why the processors and kernels cannot be configured.

Request:
1. Clarify those configurations and their update conditions
2. Need to understand if those can be applied on all the personalities: controller, worker, storage

controllers/host/memory.go
```diff
@@ -206,3 +187,46 @@ func (r *HostReconciler) ReconcileMemory(client *gophercloud.ServiceClient, inst
 
 	return nil
 }
+
+type memOptsWrapper struct {
```
yjian118, line 191: nit: suggest to name it as:
memoryWithOpt\ memoryConfig etc...

enzocandotti:
In the Test Plan, could you confirm that the sysinv logs shows no errors when updating the memory?
Should be around https://opendev.org/starlingx/config/src/commit/fb82dfc8f76f85638f944b09fc8b50911c17d60f/sysinv/sysinv/sysinv/sysinv/api/controllers/v1/memory.py#L575-L599

wasnio:
Hi, Enzo. I can confirm the platform memory was configured without any issues, but I did not pay attention to sysinv logs.

PR #477, opened 2025-08-22
Skip tasks not required after enrollment

wasnio:
This commit reduces the playbook execution time after subcloud
enrollment, removes unneeded tasks and reduces the execution time of
other tasks.

Test Cases:
- PASS: enroll a subcloud
- PASS: deploy a standard subcloud
- PASS: apply the config on an unlocked subcloud, ensure the host
        goes back to reconciled=true
- PASS: apply the config-map with the key factory-config-finalized=true
        and ensure status.reconciled=true on all the resources


hugonicodemos (approved):

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -454,20 +389,7 @@
         - wait_for:
             timeout: 5
             msg: Waiting after apply DM config
-
-        - name: Restart the deployment manager if subcloud enrollment
```
yjian118, line 458: This section was introduced in: https://github.com/Wind-River/cloud-platform-deployment-manager/commit/fcb78411ff50f87f36f40420f3dbae6347f46330
Was it no longer needed? Check the commit message: if there's no config change before and after the enrollment

yjian118 (commented):
Expect an additional test that the configuration remains the same before and after enrollment, note: need to add the routes during factory install - though not a likely case

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -266,16 +216,7 @@
                 Expected 'applied' state but the app is in
                 '{{ app_apply_status.stdout }}' state."
         when: not ("applied" in app_apply_status.stdout)
-
-    # Delete old Deployment Manager pod if it was reinstalled,
-    # the pod to delete may already be terminated by helm
-    - name: Restart Deployment Manager if reinstalled
```
yjian118, line 272: A bit concerned about removing this task:

Request test:
1. give a wrong auth password in the system endpoint in the initial deployment, check the DM was blocked by the wrong endpoint password
2. update the password and apply it again. 
The expectation is the DM can get the new password and pass the authentication. As far as I know we'll need to delete the pod once to get the new secret

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
wasnio: I did it, dm unlocked the subcloud as expected.

controllers/host/host_controller.go
```diff
@@ -1794,9 +1794,9 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
 		return nil
 	}
 
-	logHost.Info("defaults are:", "values", defaults)
-	logHost.Info("final profile is:", "values", profile)
-	logHost.Info("current config is:", "values", current)
+	// logHost.Info("defaults are:", "values", defaults)
```
yjian118, line 1797: Don't move those logs, critical for troubleshooting

docs/playbooks/wind-river-cloud-platform-deployment-manager.yaml
```diff
@@ -347,13 +280,15 @@
             (get_distributed_cloud_role.stdout is not defined) or
             get_distributed_cloud_role.stdout != "systemcontroller"
 
+          # TODO(wasnio): delete this once dm pod can handle missing webhooks
```
yjian118, line 283: Not required, but one future work item may be: update the helmchart to collaborate with the pod and the webhook

controllers/system/system_controller.go
```diff
@@ -1126,9 +1126,9 @@ func (r *SystemReconciler) ReconcileRequired(instance *starlingxv1.System, spec
 		current.Certificates = &res
 	}
 
-	logSystem.Info("spec is:", "values", spec)
+	// logSystem.Info("spec is:", "values", spec)
```
yjian118, line 1129: Same here

controllers/system/system_controller.go
```diff
@@ -1680,6 +1680,8 @@ func (r *SystemReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
+	logSystem.Info(fmt.Sprintf("XXX system is factory insalled: %t", factory))
```
yjian118, line 1684: Remove the XXX or use the system name instead?

controllers/common/common.go
```diff
@@ -194,6 +194,20 @@ func (h *ErrorHandler) HandleReconcilerError(request reconcile.Request, in error
 	cause := perrors.Cause(in)
 
 	switch cause.(type) {
+	case *gophercloud.ErrUnableToReauthenticate, gophercloud.ErrUnableToReauthenticate:
```
yjian118, line 197: In any case, we returned a pointer?

controllers/host/host_controller.go
```diff
@@ -2277,6 +2277,38 @@ func (r *HostReconciler) UpdateStatusForFactoryInstall(
 	return nil
 }
 
+func (r *HostReconciler) isDayConfigComplete(instance *starlingxv1.Host) bool {
```
yjian118, line 2280: suggest name: instanceConfigCompleted

controllers/host/host_controller.go
```diff
@@ -2277,6 +2277,38 @@ func (r *HostReconciler) UpdateStatusForFactoryInstall(
...
+	// TODO describe the reason
+	isFactory, err := r.GetFactoryInstall("deployment")
```
yjian118, line 2282: Can we avoid hard-coded namespace here? I know currently there's an issue in the playbook with the configMap, which is only set in the deployment namespace. We can add a TODO here if it need to be hard-coded.

In longer term, we should apply the configMap for factory install in the system resource's namespace(as it is requested)

controllers/host/factory.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	if len(hosts.Items) == 0 {
+		log.Info("not hosts found")
+		return IgnoreReconcile
+	}
+
+	for _, host := range hosts.Items {
+		host.Status.Reconciled = false
+		host.Status.Defaults = nil
```
yjian118, line 43: Please add a comment here to explain: set host defaults to nill for re-collecting the default post enrollment

controllers/host/host_controller.go
```diff
@@ -2228,53 +2181,27 @@ func (r *HostReconciler) UpdateConfigStatus(
...
-// UpdateStatusForFactoryInstall updates the status by checking the factory
-// install data.
-func (r *HostReconciler) UpdateStatusForFactoryInstall(
-	ns string,
-	instance *starlingxv1.Host,
-) error {
-	factory, err := r.CloudManager.GetFactoryInstall(ns)
-	if err != nil {
-		return err
+func (r *HostReconciler) instanceConfigCompleted(instance *starlingxv1.Host) bool {
```
yjian118, line 2184: Can you add a function description here? And do we still need the previous TODO?

yjian118 (changes requested):

controllers/common/factory.go
```diff
@@ -0,0 +1,44 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const FactoryInstallConfigMapName = "factory-install"
+const FactoryInstalled = "factory-installed"
+const FactoryConfigFinalized = "factory-config-finalized"
```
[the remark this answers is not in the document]
wasnio: done.

controllers/platformnetwork_controller.go
```diff
@@ -206,45 +209,38 @@ func (r *PlatformNetworkReconciler) UpdateConfigStatus(instance *starlingxv1.Pla
...
-	instance *starlingxv1.PlatformNetwork,
-) error {
-	reconciledUpdated, err := r.CloudManager.GetFactoryResourceDataUpdated(
-		ns,
-		instance.Name,
-		"reconciled",
-	)
-	if err != nil {
-		return err
+func (r *PlatformNetworkReconciler) updateStatusForFactoryInstall(ctx context.Context, obj client.Object) []reconcile.Request {
```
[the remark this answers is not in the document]
wasnio: done.

controllers/host/host_controller.go
wasnio: DONE. The old todo is not required.

controllers/host/host_controller.go
```diff
@@ -1602,36 +1603,7 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
...
 	}
 
-	// Check factory install config map
-	logHost.V(2).Info("checking factory install config map", "namespace", reqNs)
-	factory, err := r.CloudManager.GetFactoryInstall(reqNs)
-	if err != nil {
-		return perrors.Wrap(err, "failed to get factory install config map")
-	}
-
-	if factory && host.IsUnlockedEnabled() {
```
yjian118, line 1612: Can you help restore this logic? It was observed that the unlock action is very fast to kill the dm pod which causing the final state not updated

PR #488, opened 2025-11-17
Fix host availability monitoring in multi-node subclouds

wasnio:
Retrieve host count from deployment config to prevent skipping host
checks when host appear dynamically. Fail if host count from
deployment config do not appear in the enventory.

Remove the hard-coded "deployment" namespace resource monitoring task

Test Plan:
- PASS: deploy a DX subcloud with dynamic controller-1

Fail Path:
- PASS: deploy a DX subcloud with dynamic controller-1 and fail
        because controller-1 does not appear in the inventory
- PASS: deploy a DX subcloud with dynamic controller-1 and fail
        because controller-1 isn't available/online

### DIFF
--- a/docs/playbooks/roles/initial-config/tasks/monitor-deployment.yaml
+++ b/docs/playbooks/roles/initial-config/tasks/monitor-deployment.yaml
@@ -1,44 +1,61 @@
 ---
 # SPDX-License-Identifier: Apache-2.0
 # Copyright(c) 2025 Wind River Systems, Inc.
-- name: Get the number of configured hosts
-  shell: >-
-    kubectl -n {{ deployment_namespace }}
-    get hosts --no-headers | wc -l
-  environment:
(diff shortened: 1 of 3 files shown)


docs/playbooks/roles/initial-config/tasks/monitor-deployment.yaml
```diff
@@ -1,44 +1,61 @@
...
+      {{ get_host_count.stdout }} in the inventory,
+      please check the host inventory: system host-list
+  when: >-
+    get_host_count.stderr != "" or
+    configured_hosts|int > get_host_count.stdout|int
 
 - name: Wait until all hosts can be configured by dm
   shell: >-
-    source /etc/platform/openrc;
+    source /etc/platform/openrc > /dev/stderr;
```
hugonicodemos: why is this  needed?

yjian118 (approved):

PR #505, opened 2026-02-12
Log errs for each configuration apply

### LINKED ISSUE #504: DM playbook failed to apply the configuration without meaningful logs for trouble shooting
The playbook failed to apply the configuration. But the host was unlocked and rebooted during the 3rd time of the re-apply. The playbook only logs the lost of the connection, causing unable to understand the cause remotely

yjian118:
The current playbook only logs the last round of logs of applying the deployment configurations.

In the case of syntax errors in one CR, or rejected by the validation webhooks, as the host may get reconciled and unlocked, the error log about the fatal error of lost connection due to unlock may not be logged, and difficult for trouble shooting.

For the ease of trouble shooting, this commit logs the stdout and stderr after the first apply, ensuring enough logs to understand the potential problem in the config file.

This commit also updates the SETPUP_ENVTEST_VERSION to release-0.20 which fit for golang 1.23. The previous version is deprecated.

Test plan:

1. Passed - make test
2. Passed - verified the 1st apply failed with an error message,s but the retry in the rescue block continues, until the retries are exhausted.
3. Passed - verified normal apply and reconfig successful

### DIFF
--- a/docs/playbooks/roles/common/tasks/apply-config.yaml
+++ b/docs/playbooks/roles/common/tasks/apply-config.yaml
@@ -10,19 +10,30 @@
     distributed_cloud_role == "subcloud" or
     distributed_cloud_role == ""
 
-- name: Apply deployment configuration file
-  shell: kubectl apply -f {{ deploy_config }}
-  environment:
-    KUBECONFIG: "/etc/kubernetes/admin.conf"
-  register: apply_deploy_config
-  retries: 5
-  delay: 10
-  until: apply_deploy_config.rc == 0
-  failed_when: false
+- name: Apply deployment configuration with retries
+  block:
+    - name: Apply deployment configuration file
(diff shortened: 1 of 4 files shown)


docs/playbooks/roles/common/tasks/apply-config.yaml
```diff
@@ -10,19 +10,29 @@
...
+      register: apply_deploy_config
+      retries: 3
+      delay: 10
+      until: apply_deploy_config.rc == 0
+      failed_when: false
+
+    - name: Fail if all retries exhausted
+      fail:
+        msg:
+          - "Failed to apply deployment configuration after 5 attempts."
```
wasnio: suggestion: use apply_deploy_config.attempts + 1, reference: https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_loops.html.

PR #507, opened 2026-02-20
Validate IP addresses before parsing and logging improvment

wasnio:
Move IP address validation before leading zero removal in
BuildCompositeProfile to catch malformed addresses earlier.
Improve error messages to include the invalid IP value.
Add event logging when BMC credentials secret is missing.

Test Plan:
- PASS - Verify invalid IP addresses are rejected
- PASS - Validate BMC secret missing logging
- PASS - AIO-SX subcloud deploy

### DIFF
--- a/controllers/host/profile_utils.go
+++ b/controllers/host/profile_utils.go
@@ -333,6 +333,12 @@ func (r *HostReconciler) BuildCompositeProfile(host *starlingxv1.Host) (*starlin
 		composite.Interfaces.Ethernet = nil
 	}
 
+	// IP address validation must occur before profile configuration updates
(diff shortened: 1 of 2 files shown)


controllers/host/profile_utils.go
```diff
@@ -333,6 +333,10 @@ func (r *HostReconciler) BuildCompositeProfile(host *starlingxv1.Host) (*starlin
 		composite.Interfaces.Ethernet = nil
 	}
 
+	if err := r.validateProfileAddresses(host, composite); err != nil {
```
yjian118: Is the purpose to avoid error traceback by ParseIP()?

PR #510, opened 2026-02-27
Add required labels to rook-ceph examples

esilvade-wr:
For configurations that use rook-ceph, the mon, mgr, and chassis placement labels are mandatory.

If Deployment Manager is configured without the labels ceph-mgr-placement: enabled, ceph-mon-placement: enabled, and the required chassis label (for example, topology.rook.io/chassis: "group-0") defined in the deployment-config.yaml of the Host Custom Resource, the system will complete initialization and reach an online state. However, the Deployment Manager will report INSYNC=false.

These labels are required to ensure proper placement of Ceph MON and MGR pods and to maintain alignment between the Deployment Manager and the underlying Rook-Ceph topology. Without them, the cluster may appear operational, but it will remain in a degraded synchronization state from the Deployment Manager perspective.

### DIFF
--- a/examples/aio-dx-rook-ceph-deployment-controller.yaml
+++ b/examples/aio-dx-rook-ceph-deployment-controller.yaml
@@ -148,10 +148,13 @@ spec:
       port:
         name: eth1001
   labels:
+    ceph-mgr-placement: enabled
+    ceph-mon-placement: enabled
     openstack-compute-node: enabled
     openstack-control-plane: enabled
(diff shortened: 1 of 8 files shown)


wasnio:
> However, the Deployment Manager will report INSYNC=false.

AFAIK the label is applied before the reconcile process (bootstap or an app installed during bootstrap) and dm will remove the label, but the resource will go to reconcile=true & inSync=True.

PR #517, opened 2026-03-30
Refactor webhook type tests: split and consolidate

wasnio:
Remove all *_type_test.go files from internal/webhook/v1/ and reorganize their contents:

- Pure type logic tests (IsKeyEqual, DeepEqual, HasWorkerSubFunction, GetClusterName, SubFunctionFromString) moved to api/v1/types_test.go where the types are defined.

- Webhook integration tests (console validation, PtpInstance UnmarshalJSON and duplicate key rejection, PlatformNetwork validation) merged into the corresponding *_webhook_test.go files.

- CRUD-only tests (create, update label, delete) that only exercised the Kubernetes API server without testing project logic were dropped: addresspool_types_test.go, datanetwork_type_test.go, host_type_test.go, ptpinterface_type_test.go.

This gives a consistent naming convention where all webhook tests live in *_webhook_test.go and all type unit tests live in api/v1/.


yjian118 (changes requested):
Expecting a simple webook test for the ptpinterface

PR #535, opened 2026-04-24
Update host route update logic

### LINKED ISSUE #534: Duplicated host routes existed in host final profile blocking the host renconciled
It is observed that due to wrong gateway setup in the StarlingX system, the host profile had duplicated host routes:
controller.host.deployment/controller-1 defaults are: ... "routes":

{"interface":"mgmt0","subnet":"fdff:719a:bf60:1103::","prefix":64,"{}gateway":"fdff:719a:bf60:1109::1","metric":1},
{"interface":"oam0","subnet":"64:ff9b::","prefix":64,"gateway":"2620:10a:a001:d41::6:0","metric":1}
,{"interface":"mgmt0","subnet":"fdff:719a:bf60:1103::","prefix":64,"{}gateway":"fdff:719a:bf60:1109::1","metric":1}

This is blocking the host from being reconciled.
While checking the collected defaults in the metadata:
 defaults: '{"personality":"controller","administrativeState":"locked","subfunctions":["controller","worker","lowlatency"],"routes":[{"interface":"mgmt0","subnet":"fdff:719a:bf60:1103::","prefix":64,"gateway":"fdff:719a:bf60:4072::1","metric":1},
{"interface":"oam0","subnet":"64:ff9b::","prefix":64,"gateway":"2620:10a:a001:d41::6:0","metric":1}
,{"interface":"mgmt0","subnet":"fdff:719a:bf60:1103::","prefix":64,"gateway":"fdff:719a:bf60:1109::1","metric":1}]}'

The deployment manager should not dictate two routes as the same if their gateway is different.  The static routes should be treated as separate routes if the gateways differ. Let the system reject the wrong route creation instead.

yjian118:
Update host route update logic
The current host route control has multiple problems, not convenient
to use:
1. No route deletion on enabled hosts though supported by the system.
2. The default collection causing problems in multi-nodes system, that
   a faulty route configuration being persisted in the host default
   of the standby controller or worker/ storage nodes that causing
   the wrong route cannot be deleted post deployment, further, if the
   wrong route has conflicts with the correct route (e.g. subcloud's
   route to central cloud configured with a wrong gateway) the host
   can never be reconciled unless delete the host resource.
3. Routers with a different gateway are treated as same, causing
   duplicated records in the final profile after merge.

This commit updates the host route update logic:
1. avoid merging the collected default route into the final profile,
   the host routes are expected to be configured by the users
   explicitly in the configurations.
2. support stale route deletion in the enabled hosts.
3. compare routes includes the gateway, it is not realistic to have
   two routes to be configured to the same subnet on a same interface
   via two different gateways, but the system already has the
   validations that rejecting those configuration. The DM only needs
   to follow the normal logic to compare and reconcile them.

After this change, all routes are expected to be described explicitly in
the configuration file during day 1 and day 2. A drawback may cause
route deletion during day 2 operation, but the deleted routes can be
easily added back. And it also fits the design of using the deployment
manager to configure the system descriptive mode.

Test plan:
1. Normal deployment with preconfigured host routes.
2. Deploy host w/o preconfigured host, verify the existing routes
   pre-deploy deleted by the deployment manager.
3. Run day-2 operation, verify host route can be deleted w/o strategy.
4. Deploy host with two routes with only different gateway, verify the
   DM is still trying to add the routes, but rejected by the system.
5. Deploy multi-nodes subcloud with wrong route configured, verify the
   host can all be reconciled, verify the subcloud reconfig with
   deploymentScope principal can correct the routes and make subcloud
   online.

Asisited-by: kiro(auto model selection)


docs/charts/index.yaml
```diff
@@ -3,12 +3,12 @@ entries:
   wind-river-cloud-platform-deployment-manager:
   - apiVersion: v1
     appVersion: "1.0"
-    created: "2026-04-24T11:06:40.093338853-03:00"
+    created: "2026-04-27T17:27:27.549585216-04:00"
     description: A Helm chart for the Deployment Manager
-    digest: eb4ad14560aff854050b8ffe69aaf139da1e1bacdf54fe58fa3f9152fae7d164
+    digest: 78ffcced363ab9c5121db051a3e077af5c0535b7a40bab6378fc4c6ee8ae13f5
```
wasnio: why do we need changes in the helm index.yaml ?

yjian118: This commit was created before PR#533
After rebasing, I used "make helm-package HELM_FORCE=1" to force rebuild the helmchart in case any problem was caused by the rebase

wasnio (approved):

PR #536, opened 2026-05-01
Add platform TLS service parameters to DM default parameters

rahulroshan-kachchap:
The Deployment Manager filters service parameters against the DefaultParameters list in common/constants.go. Parameters not in this list are excluded from system configuration snapshots and cannot be applied during subcloud deployment.

Add tls-min-version and tls-cipher-suite (platform/config) to the DefaultParameters array so DM can export, import, and reconcile platform TLS settings. Add corresponding test case in filter_test.go to verify TLS parameters are handled as defaults.

Validation of parameter values (VersionTLS12, VersionTLS13, IANA cipher names) is performed server-side by the sysinv API.

Test Plan:
PASS - DM export from system with TLS params, TLS params appear in exported YAML.
PASS - DM deploy with TLS params in system YAML, params created on target.
PASS - DM deploy without TLS params, bootstrap defaults apply. PASS - DM reconcile, TLS params not removed as non-default. PASS - go test ./build/... --ginkgo.focus="ServiceParameterFilter"

### DIFF
--- a/common/constants.go
+++ b/common/constants.go
@@ -27,6 +27,8 @@ const ServiceParamNamePlatformMaxCpuPercentage = "cpu_max_freq_min_percentage"
 
 const ServiceParamNamePlatConfigIntelNicDriverVersion = "intel_nic_driver_version"
 const ServiceParamNamePlatConfigIntelPstate = "intel_pstate"
+const ServiceParamNamePlatformTLSMinVersion = "tls-min-version"
+const ServiceParamNamePlatformTLSCipherSuite = "tls-cipher-suite"
 const ServiceParamNameRadosgwFsSizeMB = "fs_size_mb"
 const ServiceParamNameRadosgwServiceEnabled = "service_enabled"
 const ServiceParamNameSecurityComplianceLockoutDuration = "lockout_seconds"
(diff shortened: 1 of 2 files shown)


common/constants.go
yjian118, line 2: Please update copyright

yjian118 (changes requested):
Commit content lgtm.

Missing a check:
Make tools, run the generated bin/deployctl  in a deployed system, verify the TLS params are not exported in the generated YAML

wasnio:
Commit test plan is confusing: 
> PASS - DM export from system with TLS params, TLS params appear in exported YAML.
Do you mean deployctl ? 

Also, you're adding it to the default parameters, I don't expect it to appear in the generated config.

wasnio (changes requested):
please, improve the commit message and make sure you tested deployctl, validate that deployctl generates the configuration without the new default service parameters.

rahulroshan-kachchap:
Adding TLS params to DefaultParameters causes them to be filtered OUT of deployctl exports - the ServiceParameterFilter in build/filter.go skips all DefaultParameters entries. This means TLS settings from an existing system cannot be captured and propagated to other deployments via deployctl build.

Since DM already has generic service-parameter support, TLS params placed in the System YAML spec.serviceParameters section work for apply and reconcile without any code change to DM. Adding them tovDefaultParameters is unnecessary and actually breaks the export path.
Closing this PR. DM support for platform TLS service parameters will be covered as a validation activity instead.

PR #541, opened 2026-06-15
Allow address pool reconciliation for incomplete pools

wasnio:
During Day-1, the Deployment Manager creates address pools with all controller addresses populated. When a controller host is deleted and re-added (Day-2), the address is removed from the pool but the AddressPool CR remains unchanged, so the addresspool reconciliation is not re-triggered. The pool is then treated as a "reconfiguration" and blocked by ReconcileOtherPlatformNetworksExpected.

This change adds a check in ReconcileOtherPlatformNetworksExpected to allow reconciliation when the system address pool is missing controller addresses that are defined in the AddressPool CR spec. This enables the Deployment Manager to repopulate the missing address after a host is re-provisioned.

```
Test Plan:
PASS: Deploy duplex system with DM including storage network on
      bond/VLAN. Delete controller-1 host, re-apply deployment
      config. Verify DM detects incomplete pool, updates it, and
      controller-1 reconciles successfully.
PASS: Verify that fully-configured pools are still treated as
      reconfiguration and blocked in bootstrap scope.
```

### DIFF
--- a/internal/controller/host/platformnetworks.go
+++ b/internal/controller/host/platformnetworks.go
@@ -538,6 +538,29 @@ func (r *HostReconciler) IsReconfiguration(client *gophercloud.ServiceClient, ne
 	return nil, false
 }
 
+// isAddrPoolIncomplete returns true if the system address pool is missing
+// controller addresses that are specified in the AddressPool CR spec.
+// This handles the case where a host was deleted and re-added, causing
+// its address to be removed from the pool without triggering a re-reconcile.
+func (r *HostReconciler) isAddrPoolIncomplete(client *gophercloud.ServiceClient, addrpool_instance *starlingxv1.AddressPool) bool {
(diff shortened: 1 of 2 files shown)


internal/controller/host/platformnetworks.go
```diff
@@ -538,6 +538,33 @@ func (r *HostReconciler) IsReconfiguration(client *gophercloud.ServiceClient, ne
...
+
+	if spec.Controller1Address != nil && system_addrpool.Controller1Address == "" {
+		return true
+	}
+
+	if spec.Controller0Address != nil && system_addrpool.Controller0Address == "" {
+		return true
+	}
+
+	if spec.FloatingAddress != nil && system_addrpool.FloatingAddress == "" {
```
wasnio: I'll revert this case.

enzocandotti (approved):

PR #543, opened 2026-06-23
Fix host deletion finalizer being removed before lock completes

wasnio:
When a host resource is deleted via kubectl, the controller issues a lock action to sysinv and starts a monitor to wait for the host to reach the locked/disabled state. However, the finalizer removal was deferred unconditionally, causing it to be removed even when the lock action returned an error (monitor wait). With the finalizer gone, Kubernetes garbage collects the Host CR before the controller can complete the deletion from sysinv.

This change moves the finalizer removal to only execute after ReconcileDeletedHost succeeds, ensuring the controller retains ownership of the CR until the host is fully locked and deleted from the system.

Test Plan:
```
PASS: Delete host controller-1 while unlocked. Verify host is locked,
      deleted from sysinv, and finalizer removed only after completion.
PASS: Verify a locked host can be deleted
PASS: Verify a failure to lock/delete does not remove the finalizer
      and retry works
PASS: Verify active controller host deletion is still skipped.
```

### DIFF
--- a/internal/controller/host/host_controller.go
+++ b/internal/controller/host/host_controller.go
@@ -2039,11 +2039,6 @@ func (r *HostReconciler) ReconcileResource(
 
 	if !instance.DeletionTimestamp.IsZero() {
 		if utils.ContainsString(instance.Finalizers, HostFinalizerName) {
-
-			// Remove the finalizer so we don't try to do this delete action again.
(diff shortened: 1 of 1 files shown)


wasnio:
Closes #544

enzocandotti (approved):

PR #551, opened 2026-07-17
Add a new comparison step for volume groups

### LINKED ISSUE #550: Day-2 doesn't support volume groups reconcile
In Day-2 operations, after applied a new configuration file that adds, deletes or updates the volume group section, the deployment manager is able to detect the changes but it's not able to apply them, whether in disabled or enabled state.

ejalberti:
To enable the system to compare and apply volume group changes, a new step has been added to the comparison process. This step will check for any differences in volume group configurations and ensure that they are properly handled during the comparison and application of changes.

This new step is crucial for day-2 operations.

TEST PLAN:
PASS: successfully compared volume groups with no changes PASS: successfully compared volume groups with changes PASS: successfully applied volume group changes in disabled and/or
      enabled node.
PASS: 'go vet ./... && go test ./...'

Closes: #550

### DIFF
--- a/internal/controller/host/host_controller.go
+++ b/internal/controller/host/host_controller.go
@@ -1031,6 +1031,16 @@ func (r *HostReconciler) CompareEnabledAttributes(in *starlingxv1.HostProfileSpe
 		}
 	}
 
+	if utils.IsReconcilerEnabled(utils.VolumeGroup) {
+		if (in.Storage == nil) != (other.Storage == nil) {
+			return false
+		} else if in.Storage != nil {
+			if !in.Storage.VolumeGroups.DeepEqual(&other.Storage.VolumeGroups) {
+				return false
+			}
+		}
+	}
+
 	if utils.IsReconcilerEnabled(utils.Route) {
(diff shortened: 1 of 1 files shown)


internal/controller/host/host_controller.go
```diff
@@ -1116,6 +1126,16 @@ func (r *HostReconciler) CompareDisabledAttributes(in *starlingxv1.HostProfileSp
...
+	if utils.IsReconcilerEnabled(utils.VolumeGroup) {
+		if (in.Storage == nil) != (other.Storage == nil) {
+			return false
+		} else if in.Storage != nil {
+			if !in.Storage.VolumeGroups.DeepEqual(&other.Storage.VolumeGroups) {
+				return false
+			}
+		}
+	}
+
```
wasnio: Why do we need to compare the volumes groups on both enabled & disabled attributes?

ejalberti: The user will be able to run these changes in both states, manually the configuration can be done with the host locked or not, only in runtime

wasnio (approved):

PR #553, opened 2026-07-21
Add dpll-mgr to PtpInstance CRD service enum

difuhuwr:
Add dpll-mgr to PtpInstance CRD service enum

Add dpll-mgr as a recognized service type in the PtpInstance CRD,
enabling Deployment Manager to provision dpll-mgr instances during
bootstrap via deployment YAML.

The dpll-mgr configuration is stored as a serialized JSON string
in the config_json section parameter, matching how sysinv stores
and retrieves it via the --from-file API.

Changes:
- api/v1/ptpinstance_types.go: Add dpll-mgr to kubebuilder enum
- config/crd/bases: Add dpll-mgr to CRD YAML enum
- helm chart: Regenerated with updated CRD

Test Plan:
  Unit tests: make test passes
  Lab validation (Trixie):
  PASS: Docker image builds successfully (make docker-build)
  PASS: rollout restart DM pod with built image
  PASS: CRD accepts dpll-mgr service type
  PASS: DM reconciles dpll-mgr instance into sysinv (inSync=true)
  PASS: config_json parameter stored correctly in sysinv
  PASS: CRD rejects invalid service type with validation error
  PASS: Instance removed from sysinv after CRD delete
  PASS: Full dpll-mgr JSON config reconciled end-to-end

Signed-off-by: Difu Hu <difu.hu@windriver.com>


enzocandotti (approved):

internal/webhook/v1/ptpinstance_webhook_test.go
```diff
@@ -73,6 +73,31 @@ var _ = Describe("PtpInstance webhook", func() {
...
+
+		It("should accept dpll-mgr with config_json section", func() {
+			key := types.NamespacedName{Name: "dpll-test", Namespace: "default"}
+
+			created := &starlingxv1.PtpInstance{
+				ObjectMeta: metav1.ObjectMeta{Name: "dpll-test", Namespace: "default"},
+				Spec: starlingxv1.PtpInstanceSpec{
+					Service: "dpll-mgr",
+					InstanceParameters: map[string][]string{
+						"config_json": {
```
wasnio: This config_json is noise for this change. You're only adding a new option to the service enum, validating parameters is out of scope.

If you want to cover the new option, make the test above table-driven over the accepted service values.

But honestly I don't think it's needed, the Kubernetes API already validates the enum against the generated CRD.

internal/controller/ptpinstance_controller_test.go
```diff
@@ -374,6 +374,29 @@ var _ = Describe("PtpInstance controller", func() {
 				err := reconciler.ReconcileResource(gcClient, instance)
 				Expect(err).ToNot(HaveOccurred())
 			})
+
+			It("should create a dpll-mgr instance with config_json parameter", func() {
```
wasnio: We can drop this test. The reconciler doesn't branch on service (it's passed through as-is), and the fake server always returns ptp4l, so this just repeats the existing "create new ptp instance" case with no added coverage.

wasnio (changes requested):

wasnio:
You are adding a new crd change, you have to run "HELM_FORCE=1 make" and commit helm chart.

internal/webhook/v1/ptpinstance_webhook_test.go
difuhuwr: Done. The test is removed.

internal/controller/ptpinstance_controller_test.go
difuhuwr: Done. The test is removed.

difuhuwr:
Did "HELM_FORCE=1 make" and commit helm chart

 docs/charts/index.yaml                                                |   6 +++---
 docs/charts/wind-river-cloud-platform-deployment-manager-26.03.0.tgz  | Bin 25295 -> 25298 bytes
 helm/wind-river-cloud-platform-deployment-manager/templates/crds.yaml |   1 +

wasnio (approved):

PR #559, opened 2026-07-28
Defer CGTS-VG configuration during Day-1 operation

### LINKED ISSUE #558: Alarms raised for worker Day-1 operation after LVM configuration
During Day-1 operation, compute node LVM configuration can lead to a constant delta and, consequently, alarms for unconfigured node.
```
134  INFO  192 MainThread validate_installation.verify_deployment_manager_configuration:: 
DELTA
-  "labels":  map[string]any{"lvm-csi":  "enabled"},storage:
      "volumeGroups":
-           "lvmPoolSize":         18,
+           "-part4",
```

ejalberti:
During Day-1 operation, the user can apply the lvm-csi function to the
CGTS-VG volume group. This operation results in the creation of a thin
pool using 50% of the free space. However, during compute node
bootstrapping, the CGTS-VG and other filesystems are fully configured
during the locked state, which can mistakenly lead to the creation of a
thin pool using more than the required space.

To address this issue, the LVM CSI function application is now deferred
until the DM unlocks the machine and commits the disk changes. After that,
the CGTS-VG volume group can be correctly configured.

Also, this commit includes fixes to the LVM CSI optional parameters,
such as lvmPoolSize. During configuration, the user can omit this parameter,
resulting in the default value of 50% of the free space being used for
the thin pool. However, after the configuration is applied, the deployment
manager finds a difference between the host profile and the current
host configuration, creating a cyclic false delta and system alarms.
This false delta also occurs for compute Physical Volumes that had their
device paths stripped of partition identifiers from the device path.

Closes: #558

### DIFF
--- a/api/v1/constructors.go
+++ b/api/v1/constructors.go
@@ -518,7 +518,7 @@ func parsePhysicalVolumeInfo(group *VolumeGroupInfo, vg *volumegroups.VolumeGrou
 			if partition, ok := host.FindPartition(pv.DeviceUUID); ok {
 				size := partition.Gibibytes()
 				physicalVolume.Size = &size
-				physicalVolume.Path = stripPartitionNumber(partition.DevicePath)
+				physicalVolume.Path = partition.DevicePath
 			} else {
 				msg := fmt.Sprintf("failed to lookup partition %s", pv.DeviceUUID)
 				return NewMissingSystemResource(msg)
--- a/internal/controller/host/host_controller.go
+++ b/internal/controller/host/host_controller.go
@@ -1791,6 +1791,25 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
 		return err
 	}
 
+	// Normalize volume group fields so that system-calculated lvmType and
+	// lvmPoolSize do not produce false deltas when not explicitly set by the user.
+	//
+	// This method only acts on volume group section (for lvm-csi purpose), as
+	// followed:
+	//   if one, or more, optional key is omitted by user on user profile and
+	//   the deployment manager detects the presence of a system calculated value,
+	//   the method will copy the value to the user profile, avoiding an
+	//   indesired delta and a cyclic reconciliation loop.
+	//   i.e.
+	//   volumeGroups:
+	//     - name: "cgts-vg"
+	//       lvmFunction: "lvm-csi"
+	//       <....>
+	//   System will configure the thin pool with 50% of the free space and the
+	//   default profile will detects this change. The thin pool size will be
+	//   copied to the user profile.
+	NormalizeVolumeGroupsForComparison(profile, current)
+
 	// N3000 interface name change apply
 	if host.IsUnlockedEnabled() {
 		logHost.Info("syncing interface name", "host", host.ID)
--- a/internal/controller/host/profile_utils.go
+++ b/internal/controller/host/profile_utils.go
@@ -42,17 +42,30 @@ func MergeProfiles(a, b *starlingxv1.HostProfileSpec) (*starlingxv1.HostProfileS
 	return a, nil
 }
 
-// MergeVolumeGroups keeps only VGs defined in the source profile (b),
-// using b as the source of truth.
+// MergeVolumeGroups uses b (user profile) as source of truth for which VGs
+// should exist. VGs in b that match by name in a (defaults) get their nil
+// fields filled from defaults. VGs in b with no match in a are new and
+// included as-is (to be created on the system).
 func MergeVolumeGroups(a, b *starlingxv1.ProfileStorageInfo) starlingxv1.VolumeGroupList {
 	filtered := make(starlingxv1.VolumeGroupList, 0, len(b.VolumeGroups))
 	for _, srcVG := range b.VolumeGroups {
 		for _, vg := range a.VolumeGroups {
 			if vg.Name == srcVG.Name {
-				filtered = append(filtered, srcVG)
+				if srcVG.LVMFunction == nil && vg.LVMFunction != nil {
+					srcVG.LVMFunction = vg.LVMFunction
+				}
+				if srcVG.LVMFunction != nil && *srcVG.LVMFunction != common.LVMFunctionNone {
+					if srcVG.LVMType == nil && vg.LVMType != nil {
+						srcVG.LVMType = vg.LVMType
+					}
+					if srcVG.LVMPoolSize == nil && vg.LVMPoolSize != nil {
+						srcVG.LVMPoolSize = vg.LVMPoolSize
+					}
+				}
 				break
 			}
 		}
+		filtered = append(filtered, srcVG)
 	}
 	return filtered
 }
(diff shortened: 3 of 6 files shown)


api/v1/constructors.go
```diff
@@ -518,7 +518,7 @@ func parsePhysicalVolumeInfo(group *VolumeGroupInfo, vg *volumegroups.VolumeGrou
 			if partition, ok := host.FindPartition(pv.DeviceUUID); ok {
 				size := partition.Gibibytes()
 				physicalVolume.Size = &size
-				physicalVolume.Path = stripPartitionNumber(partition.DevicePath)
```
wasnio, line 521: why we don't need stripPartitionNumber anymore?

wasnio: explain on the commit why you are removing this function usage. Also, remove the function definition, I think this function is only used here.

internal/controller/host/profile_utils.go
```diff
@@ -161,6 +174,55 @@ func FixVolumeGroupPath(a *starlingxv1.HostProfileSpec, hostInfo *v1info.HostInf
 	a.Storage.VolumeGroups = result
 }
 
+// NormalizeVolumeGroupsForComparison adjusts the profile and current VG lists
+// so that system-calculated fields (lvmType, lvmPoolSize) do not produce
+// false deltas during comparison.
+func NormalizeVolumeGroupsForComparison(profile, current *starlingxv1.HostProfileSpec, deploymentScope string) {
```
wasnio: Please, describe why we need this Normalize logic

wasnio (changes requested):

internal/controller/host/profile_utils.go
```diff
@@ -161,6 +174,55 @@ func FixVolumeGroupPath(a *starlingxv1.HostProfileSpec, hostInfo *v1info.HostInf
...
+				}
+				if currentVG.LVMPoolSize != nil && profileVG.LVMPoolSize == nil {
+					profileVG.LVMPoolSize = currentVG.LVMPoolSize
+				}
+				if currentVG.LVMFunction != nil && profileVG.LVMFunction == nil {
+					profileVG.LVMFunction = currentVG.LVMFunction
+				}
+			}
+			break
+		}
```
wasnio: Can we refactor this function into smaller functions?

wasnio (changes requested):

api/v1/constructors.go
ejalberti: Thank you for pointing this. I'll update the message commit to address this.
The strip Partition is removing the "-part<n>" from the disk path and generating a cyclic delta. The user adds the partition number and DM detects the disk without the partition.

internal/controller/host/profile_utils.go
ejalberti: Yes, of course.

internal/controller/host/profile_utils.go
ejalberti: Thank you for pointing this

api/v1/constructors.go
ejalberti: During the configuration the disk partition identificatory is important to correctly assign the PV to the CGTS-VG. This was not a problem until now because the CGTS-VG was not used on DM file on compute nodes until the lvm-csi function be created.
If the partition number or identificatory is removed and the user decide to use the lvm-csi function, the DM will create a false delta between the default value (without the identification) and the user profile (with the identificator)

ejalberti: The function is used in other contexts for parse OSD information, so the removal is not possible

internal/controller/host/profile_utils.go
ejalberti: Yes, of course. I submitted a new version

internal/controller/host/profile_utils.go
ejalberti: Thank you for the suggestion, normalize function was mentioned on commit message

api/v1/constructors.go
```diff
@@ -518,7 +518,7 @@ func parsePhysicalVolumeInfo(group *VolumeGroupInfo, vg *volumegroups.VolumeGrou
...
+				physicalVolume.Path = partition.DevicePath
```
yjian118, line 521: What's the expectation for the existing system being upgraded to the current version? The default profile is already constructed with the stripPartitionNumber.

internal/controller/host/profile_utils_test.go
```diff
@@ -1398,3 +1401,152 @@ var _ = Describe("Profile utils", func() {
...
+	Context("when user sets lvmFunction to none", func() {
+		It("should NOT inherit lvmType/lvmPoolSize from defaults", func() {
+			defaults := &starlingxv1.ProfileStorageInfo{
+				VolumeGroups: starlingxv1.VolumeGroupList{
+					{Name: "cgts-vg", LVMType: strPtr("thin"), LVMFunction: strPtr("lvm-csi"), LVMPoolSize: intPtr(35)},
+				},
+			}
+			user := &starlingxv1.ProfileStorageInfo{
+				VolumeGroups: starlingxv1.VolumeGroupList{
+					{Name: "cgts-vg", LVMFunction: strPtr("none")},
```
yjian118: Use the new constant instead

internal/controller/host/storage.go
```diff
@@ -302,6 +302,15 @@ func (r *HostReconciler) ReconcileVolumeGroups(client *gophercloud.ServiceClient
 	// Update existing volume groups capabilities and create any new volume
 	// groups as needed.
 	for _, vgInfo := range profile.Storage.VolumeGroups {
+		// Defer cgts-vg configuration until the host is enabled, this will
+		// avoid problems with free disk space
+		if vgInfo.Name == ctrlcommon.LVG_CGTS_VG &&
+			instance.Status.DeploymentScope == cloudManager.ScopeBootstrap &&
```
yjian118, line 308: Just want to confirm: is the LVG_CGTS_VG capability update allowed in day-2 operation when the host is enabled? Won't it cause issues as it on day 1?

I would assume the best place to avoid additional storage being provisioned is not on the DM side

internal/controller/host/profile_utils.go
```diff
@@ -161,6 +175,62 @@ func FixVolumeGroupPath(a *starlingxv1.HostProfileSpec, hostInfo *v1info.HostInf
...
+
+	if hasActiveFunction {
+		defaultFromCurrent(&profileVG.LVMType, currentVG.LVMType)
+		defaultFromCurrent(&profileVG.LVMPoolSize, currentVG.LVMPoolSize)
+	} else {
+		currentVG.LVMType = nil
+		currentVG.LVMPoolSize = nil
+	}
+
+	if profileVG.Name == common.LVG_CGTS_VG && deploymentScope == cloudManager.ScopeBootstrap {
```
yjian118: Same question here: why limit the scope to Bootstrap? In day2, if we want to add new capabilities, doesn't the profile need to be updated?

internal/controller/host/host_controller.go
```diff
@@ -1791,6 +1791,10 @@ func (r *HostReconciler) ReconcileExistingHost(client *gophercloud.ServiceClient
 		return err
 	}
 
+	// Normalize volume group fields so that system-calculated lvmType and
+	// lvmPoolSize do not produce false deltas when not explicitly set by the user.
+	NormalizeVolumeGroupsForComparison(profile, current, instance.Status.DeploymentScope)
```
yjian118: Need to add a description here to explain  it only applies on the scope, and why

internal/controller/host/profile_utils.go
yjian118: The bootstrap scope is overlapping the if...else block above; the currentVG may already be cleared when it is expected to be the source in this block. Is it intentional? 
The logic is weird here; can you add more docstrings for that?
If that's not intentional, can you properly design the logic?

yjian118 (changes requested):
My main concern is the backward compatiblity to already provisioned system with the default already collected, please verify day-2 operation with the change of suffix removal

internal/controller/host/storage.go
ejalberti: Yes. It's expected and allowed to the user configure the CGTS volume group with lvm-csi function. This is a runtime operation that can be executed with the system unlocked.
The configuration defer is only a secure way to configure the cgts when its already configured with the final filesystem organization.
We need to implement this approach to wait the configuration of cgts-vg be executed.

internal/controller/host/host_controller.go
ejalberti: Thank you for your suggestion. Added.

api/v1/constructors.go
ejalberti: In our tests when the system is in reconciled state (and in sync) the change of the physical volume won't change the configuration. 
After a different test, changing the deployment scope to principal and removing a label (to force a difference between the old profile)

internal/controller/host/host_controller.go
yjian118: Can you move the docstring above the actual function?

internal/controller/host/profile_utils.go
```diff
@@ -161,6 +174,62 @@ func FixVolumeGroupPath(a *starlingxv1.HostProfileSpec, hostInfo *v1info.HostInf
 	a.Storage.VolumeGroups = result
 }
 
+// NormalizeVolumeGroupsForComparison adjusts the profile and current VG lists
+// so that system-calculated fields (lvmType, lvmPoolSize) do not produce
+// false deltas during comparison.
+// During the configuration the user can omit lvmPoolSize or other keys, by
+// adding the keys in user profile DM will avoid false deltas.
+func NormalizeVolumeGroupsForComparison(profile, current *starlingxv1.HostProfileSpec, deploymentScope string) {
```
yjian118: If deploymentScope  is no longer used, please remove it

yjian118 (changes requested):

wasnio (approved):

yjian118 (approved):

PR #560, opened 2026-07-31
Add channels and sriov_vf_channels interface configuration support

wasnio:
Add support for configuring NIC channels (queues) on interfaces via
two new parameters:

- channels (pfChannels): configures the number of combined channels
  on ethernet and ae (bond) type interfaces. Must be >= 1 and <= the
  port's maxchannels.

- sriov_vf_channels (vfChannels): configures the number of SR-IOV VF
  channels on pci-sriov classed interfaces with vf-driver set to
  netdevice, or on vf type interfaces. Must be >= 1 and <= the port's
  sriov_vf_maxchannels.

```
Test Plan:
- Deploy AIO-SX/Standard system with channels configured on
  platform ethernet and pci-sriov ethernet interfaces
- Verify channels value is applied via system host-if-show
- Deploy VF interface with vfDriver=netdevice and sriov_vf_channels
  configured, verify value is applied
- Run deployctl build and confirm generated YAML includes
  pfChannels and vfChannels fields with correct values
- Verify webhook rejects sriov_vf_channels on interfaces without
  vf-driver set to netdevice
```


wasnio:
Depends on: https://github.com/Wind-River/gophercloud/pull/54

internal/webhook/v1/hostprofile_webhook.go
```diff
@@ -221,6 +221,59 @@ func validateOVSAccessInfo(obj *starlingxv1.HostProfile) error {
...
+//     'netdevice' when sriov_vf_channels is specified.
+func validateChannelsInfo(obj *starlingxv1.HostProfile) error {
+	if obj.Spec.Interfaces == nil {
+		return nil
+	}
+
+	// Validate ethernet interfaces
+	for _, eth := range obj.Spec.Interfaces.Ethernet {
+		// VFChannels on pci-sriov ethernet requires vf-driver netdevice
+		if eth.VFChannels != nil && eth.Class == "pci-sriov" {
```
yjian118, line 237: if eth.Class != "pci-sriov" do we need to return an error?

yjian118 (commented):
What if we create a SR-IOV w/o the vfChannels configuration? Will it block the interface update? If the sysinv  block it when interface update, then I'm afraid we'll need to handle it in the dm or update existing configuration to avoid massive failure in the labs

internal/webhook/v1/hostprofile_webhook.go
wasnio: No, we should accept the parameters on all kinds of interface. But sr-iov and vf, we should ensure that the user also configures the vfdriver as netdevice.

wasnio:
@yjian118 if channel isn't in the configuration, dm won't send it to sysinv.

yjian118 (approved):

enzocandotti (approved):

PR #561, opened 2026-08-11
Fix VIM strategy permanently failing on subsequent Day-2 operations

wasnio:
The VIM client cached by the Deployment Manager becomes permanently unusable after the first Day-2 strategy completes, blocking all subsequent Day-2 operations that require lock/unlock orchestration.

This commit resets the client when the token expires, dm will create a new one during the subsequent reconcile loop.

```
Test Plan:
PASS: Deploy AIO-DX, apply Day-2 config change on controller-0
      (DM running on controller-1), verify strategy completes
PASS: Wait >1 hour (token expiry), apply second Day-2 config
      change, verify DM resets VIM client and strategy succeeds
```

### DIFF
--- a/internal/controller/manager/manager.go
+++ b/internal/controller/manager/manager.go
@@ -868,7 +868,9 @@ func (m *PlatformManager) ClearStrategy() {
 	if err != nil {
 		log.Error(err, "Set strategy applied sent false error")
 	}
+	namespace := m.strategyStatus.Namespace
 	m.strategyStatus = NewStrategyStatus()
+	m.strategyStatus.Namespace = namespace
 
 	// Reset strategy retry count
 	err = m.SetStrategyRetryCount(0)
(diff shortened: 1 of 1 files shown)


internal/controller/manager/manager.go
yjian118: I think you'll need it for delete as well

internal/controller/manager/manager.go
```diff
@@ -958,12 +974,20 @@ func (m *PlatformManager) GcDelete(c *gophercloud.ServiceClient) (r systemconfig
 
 // GcActionStrategy is wrapper function for systemconfigupdate ActionStrategy
 func (m *PlatformManager) GcActionStrategy(c *gophercloud.ServiceClient, opts systemconfigupdate.StrategyActionOpts) (*systemconfigupdate.SystemConfigUpdate, error) {
-	return systemconfigupdate.ActionStrategy(c, opts)
+	result, err := systemconfigupdate.ActionStrategy(c, opts)
+	if isVimAuthError(err) {
+		m.ResetVimClient()
+	}
```
yjian118: Why not execute the op again after resetting the vim client? Same comment on other Gc* wrapper

internal/controller/manager/monitor.go
```diff
@@ -278,6 +293,9 @@ func monitorStrategyState(management CloudManager) bool {
 		_, err = management.GcActionStrategy(client, action)
 		if err != nil {
 			log.Error(err, "Strategy apply failed.")
+			if isVimAuthError(err) {
```
yjian118: There's no need to update this method if GcActionStrategy retries after refresh the vim client

internal/controller/manager/monitor.go
```diff
@@ -487,6 +505,9 @@ func ManageStrategy(management CloudManager) bool {
 			management.SetStrategyExpectedByOtherReconcilers(false)
 			if err != nil {
 				log.Error(err, "Strategy creation failed")
+				if isVimAuthError(err) {
```
yjian118: Same here

yjian118 (changes requested):

yjian118 (approved):

enzocandotti (approved):

PR #563, opened 2026-08-17
Turns physicalVolumes nullable for CGTS-VG

### LINKED ISSUE #564: Turns physicalVolumes nullable for CGTS-VG
During LVM configuration it's necessary to inform physical volumes assigned for CGTS volume group. CGTS-VG is normally assigned to a partition, what requires not only the device path but the partition size. During the configuration for day-1 the partition size is available for controller-0 but not necessarily available for the other hosts. Other important situation is that the size needs to be precisely informed, otherwise the reconcile function can try to resize the volume group/partition, leading to an error.

Turning the physical volumes nullable, only for CGTS-VG, can improve the configuration experience and avoid errors during the day-1 operation.

ejalberti:
To avoid undersiderable changes on CGTS-VG physical volumes, like the partition size or device path, this change turns physicalVolumes section nullable for CGTS-VG only.

For any other volume group kubernetes XValidation will raise an error for the absence of physicalVolumes section.

TEST PLAN:
PASS: successfully assigned lvm-csi function to CGTS-VG without PV
PASS: successfully assigned lvm-csi function to CGTS-VG with PV
PASS: successfully rejected any attempt to configure non CGTS-VG volume
      group without PV information
      
Closes: #564


yjian118:
Can you help to add examples about this configuration?

api/v1/hostprofile_types.go
```diff
@@ -225,6 +225,7 @@ type PhysicalVolumeList []PhysicalVolumeInfo
 // +deepequal-gen:ignore-nil-fields=true
 // +kubebuilder:validation:XValidation:rule="!has(self.lvmPoolSize) || (has(self.lvmType) && self.lvmType == 'thin')",message="lvmPoolSize can only be set when lvmType is 'thin'"
 // +kubebuilder:validation:XValidation:rule="!has(self.lvmType) || (has(self.lvmFunction) && self.lvmFunction == 'lvm-csi')",message="lvmType can only be set when lvmFunction is 'lvm-csi'"
+// +kubebuilder:validation:XValidation:rule="self.name == 'cgts-vg' || (has(self.physicalVolumes) && size(self.physicalVolumes) > 0)",message="physicalVolumes is required unless the volume group name is 'cgts-vg'"
```
wasnio, line 228: It’s not a blocker, but I prefer having these validations in the webhook so we can create test cases

wasnio (approved):

api/v1/hostprofile_types.go
ejalberti: Using XValidation we can simplify the code changing. The result of this change does not affect the tests.

yjian118:
Regarding the examples, can you follow the existing examples with rook-ceph, and use the kustomize to generated it based on the template?

examples/aio-dx/lvm/lvm-storage.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: starlingx.windriver.com/v1
+kind: HostProfile
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: controller-profile
+  namespace: deployment
+spec:
+  labels:
+    openstack-control-plane: null
```
yjian118: Is the openstack-control-plane: null, openstack-compute-node: null, openvswitch: enabled and sriov: enabled requested for this configuration? If not, please remove those lines to make the patch clear

examples/aio-sx/lvm/lvm-storage.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: starlingx.windriver.com/v1
+kind: HostProfile
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: controller-profile
+  namespace: deployment
+spec:
+  labels:
+    openstack-control-plane: null
```
yjian118: Same comment on this file

yjian118 (approved):

PR #566, opened 2026-09-03
Avoid reconcile when CGTS-VG is absent in user profile

### LINKED ISSUE #565: Host cannot reach INSYNC=True if cgts-vg is not specified in the deployment configuration when configure lvm_type thin
When only dedicated volume group is used to support lvm-csi function, without offering the cgts-vg volume group, the host cannot reach INSYNC true

ejalberti:
During host configuration, the user may set an lvm-csi function in one or more volume groups (VGs) in dedicated mode (when a specific disk is assigned to the VG). In this configuration, CGTS-VG (the volume group that supports the system) can be omitted.

It is expected that this configuration is successfully applied and the host reaches the 'INSYNC=true' state. However, omitting CGTS-VG creates a constant delta caused by the CGTS-VG deletion protection method.

This commit adds a new comparison method that prevents reconciliation, and consequently the 'INSYNC=false' status when the only difference between the host and the user profile is the presence of CGTS-VG in the host's volume group list.

TEST PLAN:
PASS: Node successfully reaches INSYNC=true when configured with only
      CGTS-VG for the lvm-csi function.
PASS: Node successfully reaches INSYNC=true when configured with only a
      dedicated volume group (in thin or thick mode) for the lvm-csi
      function.
PASS: Node successfully reaches INSYNC=true when configured with both
      CGTS-VG and a dedicated volume group for the lvm-csi function.

Closes: #565

### DIFF
--- a/internal/controller/host/host_controller.go
+++ b/internal/controller/host/host_controller.go
@@ -961,6 +961,32 @@ func (r *HostReconciler) CompareAttributes(in *starlingxv1.HostProfileSpec, othe
 		r.CompareDisabledAttributes(in, other, instance.Namespace, personality, false)
 }
 
+// volumeGroupsEqualIgnoringCurrentCGTS reports whether the desired volume group
+// list (in) and the current volume group list (other) are equivalent once a
+// "cgts-vg" that is present only in the current config is disregarded.
+func volumeGroupsEqualIgnoringCurrentCGTS(in, other *starlingxv1.VolumeGroupList) bool {
+	// If the user profile declares "cgts-vg" then there is nothing special to
+	// ignore; fall back to the regular comparison.
+	for i := range *in {
+		if (*in)[i].Name == common.LVG_CGTS_VG {
+			return in.DeepEqual(other)
+		}
+	}
+
+	// Build a copy of "other" without the current-only "cgts-vg" entry.
+	// If there is no such entry to drop the copy is identical to "other",
+	// so this also handles the case where the lists are already equal.
+	filtered := make(starlingxv1.VolumeGroupList, 0, len(*other))
+	for i := range *other {
+		if (*other)[i].Name == common.LVG_CGTS_VG {
+			continue
+		}
+		filtered = append(filtered, (*other)[i])
+	}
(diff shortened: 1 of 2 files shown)


internal/controller/host/host_controller.go
```diff
@@ -1035,7 +1063,12 @@ func (r *HostReconciler) CompareEnabledAttributes(in *starlingxv1.HostProfileSpe
 		if (in.Storage == nil) != (other.Storage == nil) {
 			return false
 		} else if in.Storage != nil {
-			if !in.Storage.VolumeGroups.DeepEqual(&other.Storage.VolumeGroups) {
+			// Compare volume groups section. If VGTS-VG volume group is only
```
yjian118: CGTS-VG

internal/controller/host/host_controller.go
```diff
@@ -961,6 +961,34 @@ func (r *HostReconciler) CompareAttributes(in *starlingxv1.HostProfileSpec, othe
...
+		if (*in)[i].Name == common.LVG_CGTS_VG {
+			return in.DeepEqual(other)
+		}
+	}
+
+	// Build a copy of "other" without the current-only "cgts-vg" entry.
+	// If there is no such entry to drop the copy is identical to "other",
+	// so this also handles the case where the lists are already equal.
+	filtered := make(starlingxv1.VolumeGroupList, 0, len(*other))
+	removed := false
```
yjian118: Why to have this removed flag? Expecting multiple records of CGTS_VG? Otherwise, it is not necessary
filtered := make(starlingxv1.VolumeGroupList, 0, len(*other))
for i := range *other {
    if (*other)[i].Name == common.LVG_CGTS_VG {
        continue
    }
    filtered = append(filtered, (*other)[i])
}

yjian118 (changes requested):
Some minor change expected

internal/controller/host/host_controller.go
ejalberti: Sorry

yjian118 (approved):

PR #568, opened 2026-09-10
Add synce4l to PtpInstance CRD service enum

cbrwr:
Add synce4l as a recognized service type in the PtpInstance CRD, enabling Deployment Manager to provision synce4l instances during bootstrap via deployment YAML.

Test Plan:
PASS: Docker image builds successfully (make docker-build) 
PASS: rollout restart DM pod with built image
PASS: CRD accepts synce4l service type
PASS: DM reconciles synce4l instance into sysinv (inSync=true) 
PASS: synce4l correctly generated in /etc/linuxptp/ptpinstance


wasnio (approved):

yjian118 (approved):

yjian118:
@wasnio @cbrwr we may need to update the image tag integrated in the build, if we need to verify the result after this change

PR #569, opened 2026-09-21
Apply interface channels at runtime without a host lock

### LINKED ISSUE #570: day-2 operations do not support changing NIC channels (queues) at runtime
Commit #560 : https://github.com/Wind-River/cloud-platform-deployment-manager/pull/560  added support for specifying the PF and VF channels for an interface.  

Currently, all DM interface operations require a host lock/unlock.  However, channels and sriov_vf_channels can be done at runtime, without a host lock/unlock.  

For channel operations, it should be allowed via a day-2 override to change the channel config without a host lock/unlock

steven-webster:
Builds on #560 (channels/sriov_vf_channels schema, validation and deployctl support) by reconciling channel changes at runtime.

pfChannels and sriov_vf_channels are runtime-applyable without a host lock (a patch limited to /channels or /sriov_vf_channels sets check_host=False). Reconcile them on the enabled (unlocked) host via a new ReconcileInterfaceChannels step, and classify a channel-only delta as an enabled (runtime) attribute so it is applied without forcing an unnecessary lock/unlock. Non-channel interface differences continue to require a lock as before.

An invalid value (for example sriov_vf_channels greater than the port's sriov_vf_maxchannels for the current VF count) is rejected by sysinv; the error is reported through the standard reconcile error path (logged, host reported not-reconciled) and the operator corrects the profile.

Test Plan (AIO-DX):
- Day-2 set pfChannels on a platform ethernet interface (cluster0): applied at runtime on both controllers via system host-if-show, no host lock, host remains reconciled.
- Day-2 set a valid sriov_vf_channels (e.g. 2) on a pci-sriov netdevice interface: applied at runtime on both controllers, no lock.
- Day-2 set sriov_vf_channels above the port maximum (16 at 32 VFs, max 4): value not applied (stays at prior value), no lock, and the sysinv error "The requested vf-channels (N) exceeds the maximum supported (M) ..." appears in the deployment-manager log; host is reported not-reconciled until corrected.
- Non-channel interface change still triggers the expected lock/unlock.
- go test ./internal/controller/host/ passes.


wasnio (approved):

examples/aio-sx-nic-channels.yaml
```diff
@@ -0,0 +1,224 @@
+apiVersion: v1
```
yjian118, line 1: examples are expected to have a base ref and make patch on top of it with kustomization, can you update it accordingly, check in aio-sx/single-nic

yjian118 (changes requested):

examples/aio-sx-nic-channels.yaml
steven-webster: I have restructured the pull request to include the kustomization patch and generated the flat example with make examples

yjian118 (approved):

